PR #47, opened 2016-07-04
PAYARA-835 Documentation for logging server log entries to a file

MeroRai:


mikecroft:
Doesn't this also expose a programmatic way to set the log file? Can you also explain a bit about how it works and give an example of how to use it? Do I need to specify a log file name, or if I don't specify an argument, will it create a default named file or will it fail? If there is a default, what is it called? If I supply a filename without a path, where is the file created?

PR #49, opened 2016-07-12
Added custom JACC Provider registration instruccions

ggam:
Simple instructions but hopefully enough to help users get some insight.

Based on the research from @arjantijms.

PR #57, opened 2016-07-27
162

ggam:
Follow up from #49 targeting 162 branch and providing CLI guidelines (thanks to the great asadmin recorder).


mikecroft:
Manually merged

PR #63, opened 2016-08-05
Updated Healthcheck service docs

fturizo:
Detailed configuration on missing metrics, expanded parameter information for asadmin commands, clarified how some metrics must be configured.


mikecroft:
In the process of manually merging

PR #65, opened 2016-08-08
Request tracing and notification services

mikecroft:


mikecroft:
@mulderbaba is double checking the asadmin commands to make sure I have everything correct. I'll wait for his approval before merging

PR #79, opened 2016-10-25
Added the deployDir alias to the same row as deploymentDir

mikecroft:


mikecroft:
incorrectly set PR to master

PR #82, opened 2016-10-28
164 new features - notification and request tracing

OndrejM:
PYSVC-123 - Added docs about notification service enabled by default (preview [here](https://ondrejm.gitbooks.io/payara-server-documentation-fork/content/v/130e9567eed24444eb110480d15760fbaaba14aa/documentation/extended-documentation/notification-service/notification-service.html))

Added docs to request tracing. List of events and requests traced (preview [here](https://ondrejm.gitbooks.io/payara-server-documentation-fork/content/v/130e9567eed24444eb110480d15760fbaaba14aa/documentation/extended-documentation/request-tracing-service/request-tracing-service.html))

### DIFF
--- a/documentation/extended-documentation/notification-service/notification-service.md
+++ b/documentation/extended-documentation/notification-service/notification-service.md
@@ -2,7 +2,7 @@
 
 Payara Server version 4.1.1.163 introduced a general Notification service in order to log events which come from other services, such as the [Health Check service](/documentation/extended-documentation/health-check-service/health-check-service.md) or the [Request Tracing service](/documentation/extended-documentation/request-tracing-service/request-tracing-service.md)
 
-The Notification service provides disseminating of notification events through various channels that are being created by Request Tracing Services and/or Health Check Services. As of {{ book.currentVersion }}, the Notification service integrates with log notification mechanism which can be seen in the `domain.xml`:
+The Notification service provides dissemination of notification events through various channels that are being created by Request Tracing Services and/or Health Check Services. As of the version 4.1.1.164, the Notification service integrates with log notification mechanism which can be seen in the `domain.xml`:
 
 ```
 <notification-service-configuration enabled="true">
@@ -13,7 +13,7 @@ The Notification service provides disseminating of notification events through v
 
 The main configuration tag `notification-service-configuration`  has only one attribute named `enabled`, which can be set to either `true` or `false`. This enables or disables the whole notification distributing.
 
-The `<log-notifier-configuration>` tag registers a log notifier to the pub-sub model of notification service. The notifier subscribes to each log notification event that would get published either by health check services or request tracing services, and writes information about the event into the server log file.
+The `<log-notifier-configuration>` tag registers a log notifier to the pub-sub model of the notification service. The notifier subscribes to each log notification event that would get published either by health check services or request tracing services, and writes information about the event into the server log file.
 
 By default, each Payara Server domain is configured with log notifier enabled in the domain.xml, by setting `enabed="true"` for both `notification-service-configuration` and `log-notifier-configuration`. This is the recommended setup until more notifiers are available, as the default value of `enabled` is false, which means that notifications are ignored.
 
--- a/documentation/extended-documentation/notification-service/notification-service.md
+++ b/documentation/extended-documentation/notification-service/notification-service.md
@@ -15,7 +15,8 @@ The main configuration tag `notification-service-configuration`  has only one at
 
 The `<log-notifier-configuration>` tag registers a log notifier to the pub-sub model of the notification service. The notifier subscribes to each log notification event that would get published either by health check services or request tracing services, and writes information about the event into the server log file.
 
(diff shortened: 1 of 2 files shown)


documentation/extended-documentation/request-tracing-service/request-tracing-service.md
```diff
@@ -6,5 +6,26 @@ The Request Tracing Service provides tracing facilities for Servlet request even
...
+
+ - start/end of the request
+ - an EJB method entered/exited
+ - a CDI bean method entered/exited (only configured methods are observed)
+ - SQL statement invocations via JDBC
+ - a JTA transaction is started/finished
+ - JASPIC authentication
+
+
 * [Asadmin Commands](asadmin-commands.md)
```
mikecroft: It might be good to add a horizontal rule here to add a bit of separation between the last section and these reference links, like the bottom of the user guides:

https://github.com/payara/Payara-Server-Documentation/blob/next/documentation/user-guides/upgrade-payara.md

documentation/extended-documentation/notification-service/notification-service.md
```diff
@@ -2,7 +2,7 @@
 
 Payara Server version 4.1.1.163 introduced a general Notification service in order to log events which come from other services, such as the [Health Check service](/documentation/extended-documentation/health-check-service/health-check-service.md) or the [Request Tracing service](/documentation/extended-documentation/request-tracing-service/request-tracing-service.md)
 
-The Notification service provide disseminating of notification events through various channels that are being created by Request Tracing Services and/or Health Check Services. As of {{ book.currentVersion }}, the Notification service integrates with log notification mechanism which can be seen in the `domain.xml`:
+The Notification service provides disseminating of notification events through various channels that are being created by Request Tracing Services and/or Health Check Services. As of {{ book.currentVersion }}, the Notification service integrates with log notification mechanism which can be seen in the `domain.xml`:
```
mikecroft: "...provides dissemination of notification events..."

Would it be safer to explicitly say "As of 4.1.1.164" rather than using the `currentVersion`, since that will change with each release? Or just remove the version reference, since all future versions will integrate with the log mechanism.

documentation/extended-documentation/notification-service/notification-service.md
```diff
@@ -11,14 +11,16 @@ The Notification service provide disseminating of notification events through va
 ```
 
 
-The main configuration tag `notification-service-configuration` which has only one attribute named `enabled`, which can be set to either `true` or `false`. and this enables or disables the whole notification distributing.
+The main configuration tag `notification-service-configuration`  has only one attribute named `enabled`, which can be set to either `true` or `false`. This enables or disables the whole notification distributing.
 
-The `<log-notifier-configuration>` tag registers a log notifier to the pub-sub model of notification service where it subscribes to each log notification event that would get published either by health check services or request tracing services.
+The `<log-notifier-configuration>` tag registers a log notifier to the pub-sub model of notification service. The notifier subscribes to each log notification event that would get published either by health check services or request tracing services, and writes information about the event into the server log file.
```
mikecroft: "...to the pub-sub model of **_the_** notification service..."

documentation/extended-documentation/notification-service/notification-service.md
```diff
@@ -11,14 +11,16 @@ The Notification service provide disseminating of notification events through va
...
+
+By default, each Payara Server domain is configured with log notifier enabled in the domain.xml, by setting `enabed="true"` for both `notification-service-configuration` and `log-notifier-configuration`. This is the recommended setup until more notifiers are available, as the default value of `enabled` is false, which means that notifications are ignored.
```
mikecroft: This isn't too clear - is the log notifier enabled or disabled by default? It seems like you're saying that it is enabled by default, but you need to explicitly set the value of enabled to true, meaning it's not enabled by default?

documentation/extended-documentation/notification-service/notification-service.md
OndrejM: Ok, I'll change it. I didn't write the paragraph, just corrected a mistake. What you suggest makes sense and I will update it acoordingly.

documentation/extended-documentation/notification-service/notification-service.md
OndrejM: You are right. The fact is that what was changed in 164 is only the configuration in the domain. If people upgrade from 163, it will still be disabled, as the default for enable is false in both cases.

I don't know how to write it clearly.

OndrejM:
I fixed all but the last comment. I'll try to improve the wording about the enabled log-notifier, but currently I don't know to do it.

OndrejM:
I tried to reword the paragraph. I hope that it is more clear now.

documentation/extended-documentation/notification-service/notification-service.md
OndrejM: I tried to reword the paragraph. I hope that it is more clear now.

mikecroft:
Great, that's much clearer!

PR #98, opened 2016-12-22
Maven config for running Payara Micro and packaging uber JAR

OndrejM:


documentation/payara-micro/maven/maven.md
```diff
@@ -8,6 +11,88 @@ In your project's POM, include the following dependency:
 <dependency>
     <groupId>fish.payara.extras</groupId>
     <artifactId>payara-micro</artifactId>
-    <version>4.1.1.162</version>
+    <version>{{ book.currentVersion }}</version>
```
mikecroft: Nice idea! I would never have thought to use that variable here :-)

OndrejM: It doesn't seem to work though, I forgot to check it on gitbooks. I'll revert the version number.

OndrejM: I found a better solution: #99

PR #105, opened 2016-12-29
Security Page details, PYSVC-153

fturizo:
Includes a new article page with details on how to search reported _CVEs_ on Payara Server and how to report security issues. @smillidge, please review is this is enough for the request on **PYSVC-153**


security.md
```diff
@@ -0,0 +1,34 @@
...
+* [3. Locating Security Vulnerabilities](#3-locating-security-vulnerabilities)
+* [4. Reporting Security Issues](#4-reporting-security-issues)
+
+## 1. Overview
+
+This section describes how to deal with known or unknown security issues and common vulnerabilities and exploits \(CVEs\) related to Payara Server and Payara Micro.
+
+## 2. Download Security Fixes
+
+Reported security vulnerabilities by the community or Payara Services Limited's costumers are patched and released either in specific patch releases \(for paid support customers exclusively\) or quarterly releases. To get the specific fix for a reported vulnerability, please download the specific release that fixed that vulnerability.
```
mikecroft: Might be worth also mentioning that we provide hotfix builds to the community immediately for very serious bugs. It's only happened once before:
http://blog.payara.fish/arbitrary-file-read-exploit-hotfix

security.md
```diff
@@ -0,0 +1,34 @@
...
+
+For each of these security vulnerabilities, a reference to the _**Pull Requests**_ merged to fix them is given for auditing purposes. If a security vulnerability has been reported, but is not related or it doesn't affect Payara Server or Payara Micro in any way, a proper justification will be detailed in the observations of such vulnerability. This is done to track reported vulnerabilities and keep users informed about them.
+
+## 4. Reporting Security Issues
+
+Payara Services Limited is very active at identifying and fixing possible security vulnerabilities included into Payara Server and Payara Micro that are either inherited from GlassFish upstream or introduced by new features developed.
+
+We strongly encourage users to report such problems in the following ways:
+
+1. If you are a customer of support services, create a ticket in the [support portal](http://support.payara.fish) describing the security vulnerabilities detected as you would do for any other bug reports.
```
mikecroft: I'm not sure about the mentions of Payara Services Ltd in this section...I think the docs should probably be kept free of commercial involvement, so they are focused solely on the product rather than on support services. Happy to be contradicted on this, though!

what do you think, @OndrejM ; @fturizo ; @smillidge ?

mikecroft: This might be an exception, though, thinking about it...since it's about reporting security issues so it kind of makes sense to include it

OndrejM: I think that if we change the sentence a bit to be more general and not provide direct link to our support portal, it would make more sense to me.

I suggest this:

> If you have a support contract, create a ticket for your support provider, describing the security vulnerabilities detected as you would do for any other bug reports.

fturizo: Yeah I agree that making specific mentions to the support portal and the company maybe not the best choice since we want the docs be community centered. @OndrejM's comment is on point. I'll make the change as requested.

security.md
fturizo: Nice catch. Thanks!

PR #108, opened 2017-01-16
PYSVC-58

michaelranaldo:


michaelranaldo:
Issues with other commits caught in crossfire

PR #113, opened 2017-01-20
Payara Public API Documentation (PYSVC-248) 

fturizo:
Wrote article page for new Public API included on release 171.


mikecroft:
Added @Pandrex247 as a reviewer since he'll know more of the capabilities of the API

mikecroft:
I think this depends on this PR which hasn't been merged yet:
https://github.com/payara/Payara/pull/1211

So even if this is OK, I would wait before merging

PR #122, opened 2017-02-12
PYSVC-149 CDI Event Bus Changes in 171

OndrejM:
Preview here: https://ondrejm.gitbooks.io/payara-server-documentation-fork/content/v/921436b8c22bdd13f80e6b9d1ab28ca9ffa3ee60/documentation/extended-documentation/cdi-events.html


documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+```
+public void receiveEvent(@Observes @Inbound String message) {
+}
+```
+
+The observers may reside in any application running on one or more instances in a Hazelcast cluster, and they will listen to events fired in an application running on any instance within the same cluster. By default, events will be fired on all remote cluster instances (all instances except the one, in which the event was fired).
+
+An event will be matched with observers by the payload class, using the same rules as usual CDI events. Observers will receive events that are assignable to the observed class.
+
+Note: Additional qualifiers are not supported. The event will always be fired with no qualifiers other than `@Inbound`. Further filtering of events is possible using an [event name](#event-name).
```
fturizo: I'd change this beginning of the phrase to **IMPORTANT**

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+Several instance names are configured as an array:
+
+```
+@Inject
+@Outbound(instanceName = {"Instance-1", "Instance-2"})
+Event<String> event;
+```
+
+Instance name of Payara DAS is its name, `server` by default. Instance name of a standalone instance is its instance name. Instance name of a Payara Micro instance is the name provided the `--name` command line argument, or it's generated automatically.
```
fturizo: I'd rephrase this for more clarity,  like: 

"If you are using Payara Server, the name of the DAS' instance is server by default. In the case of a separate standalone instance, use the instance's name in the configuration. If you are using Payara Micro, the instance's name is configured using the --name option, or is generated by default if not configured."

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Loopback
+
+When an outbound event is fired, by default, it is treated as a remote message and will be delivered only to remote instances in the cluster. It will not be delivered to listeners on the same server instance.
+
+If it should be delivered also to local observers, it can be marked with `loopBack` boolean attribute:
```
fturizo: I think this would read better like this:

"If this event is needed to be delivered to observers in the local instance as well, it can be done by setting the loopBack attribute to true:"

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+@Inject
+@Outbound(loopBack = true)
+Event<String> event;
+```
+
+Such event will be delivered to observers on all instances in the cluster, including the local instance.
```
fturizo: I'd remove this line : 

"Such event will be delivered to observers on all instances in the cluster, including the local instance."

Since it's a bit redundant with the previous explanation

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+Payara Server is able to listen to and fire CDI events across the Hazelcast cluster. The CDI events can be fired and listened to by applications on Payara Server instances as well as Payara Micro instances provided that they are connected to the same cluster.
+
+This feature is not yet available in Payara Server Web Profile.
+
+Since: 4.1.1.171 (Payara Server), 4.1.153 (Payara Micro)
+
+## Initializing the event bus
+
+The event bus is initialized automatically after an application is started. No additional configuration is necessary to fire and listen to remote events.
```
OndrejM: I was thinking about it, but then I forgot it. There is still the initialize() method, so I'll add a short section about it.

OndrejM: Oh, I know why I didn't mention it - in previous releases, the CDI bus was only available in Payara Micro. I'll mention the manual initialization in the Payara Micro specific section: https://github.com/payara/Payara-Server-Documentation/pull/122/files#diff-163c05391d82d5f89649e7c560ee3138

OndrejM: In fact, I mentioned it in the Payara Micro specific docs already - see the section Manual initialization of the event bus in https://github.com/payara/Payara-Server-Documentation/pull/122/files#diff-163c05391d82d5f89649e7c560ee3138 :)

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+The payload passed as an argument to the `fire()` method has to implement the `Serializable` interface.
+
+An outbound event will be fired asynchronously on a remote instance. If [loopback](#loopback) is allowed, it will be fired also on the same instance, but in a different thread. Therefore the call doesn't block the current thread.
+
+Note: It is not possible to use additional qualifiers to match event with observers. Any additional qualifier other than `@Outbound` will mean that the event will be ignored.
+
+## Observing remote CDI events
+
+Remote events are received by beans, which observe CDI events that are marked by the `@Inbound` qualifier annotation:
```
OndrejM: I agree, I'll change it.

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+public void receiveEvent(@Observes @Inbound String message) {
+}
+```
+
+The observers may reside in any application running on one or more instances in a Hazelcast cluster, and they will listen to events fired in an application running on any instance within the same cluster. By default, events will be fired on all remote cluster instances (all instances except the one, in which the event was fired).
```
OndrejM: ok

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,136 @@
...
+
+```
+    event.fire("This is an event");
+```
+
+The payload passed as an argument to the `fire()` method has to implement the `Serializable` interface.
+
+An outbound event will be fired asynchronously on a remote instance. If [loopback](#loopback) is allowed, it will be fired also on the same instance, but in a different thread. Therefore the call doesn't block the current thread.
+
+Note: It is not possible to use additional qualifiers to match event with observers. Any additional qualifier other than `@Outbound` will mean that the event will be ignored.
```
OndrejM: I didn't want to drive so much attention to it. What about using NOTE: in capitals instead?

documentation/extended-documentation/cdi-events.md
OndrejM: Ok, it reads better.

documentation/extended-documentation/cdi-events.md
OndrejM: ok

documentation/extended-documentation/cdi-events.md
OndrejM: ok

documentation/extended-documentation/cdi-events.md
fturizo: Ah, makes sense. Then it's fine to leave it like this.

documentation/extended-documentation/cdi-events.md
fturizo: That's the point: It's important since there are use cases where you would combine qualifiers or use an stereotype and it's important to let developers know that they aren't supported.

documentation/extended-documentation/cdi-events.md
```diff
@@ -0,0 +1,134 @@
...
+
+Several instance names are configured as an array:
+
+```
+@Inject
+@Outbound(instanceName = {"Instance-1", "Instance-2"})
+Event<String> event;
+```
+
+If you are using Payara Server, the name of the DAS instance is `server` by default. In the case of a separate standalone instance, use the instance's name in the configuration. If you are using Payara Micro, the instance's name is configured using the --name option, or is generated by if not configured.
```
OndrejM: A reminder to add a link to instance-names.md once PR #127 is merged.

PR #124, opened 2017-02-14
PYSVC-315 env variable substitution in domain xml

OndrejM:
Preview: https://ondrejm.gitbooks.io/payara-server-documentation-fork/content/v/6d1ac7b9a500d61443fbdb2d8261ffa717553f50/documentation/core-documentation/configuration/configuration.html


documentation/core-documentation/configuration/configuration.md
```diff
@@ -0,0 +1,3 @@
+# Payara Server configuration files
+
+The documentation about Payara server configuration files.
```
michaelranaldo: Nitpicking - core-documentation.md follows the pattern of not having the "the"?

documentation/core-documentation/configuration/var-substitution.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+The system properties can be specified by
+
+ - the JVM (standard JVM properties like `java.home`)
+ - Payara Server (a set of predefined variables, such as `com.sun.aas.instanceRoot`)
+ - the `create-system-properties` asadmin command (or in the Admin console `System properties` section)
+ - a JVM option prefixed with `-D`
+
+When custom properties specified are specified, their value can also contain other variable references. It is thus, for example, possible to define a system property using environment variables.
```
michaelranaldo: "When custom properties are specified, their value can also contain other variable references. It is therefore possible to define a system property using environment variables."

documentation/core-documentation/configuration/var-substitution.md
```diff
@@ -0,0 +1,45 @@
...
+
+## References to system properties
+
+Variable references to system properties start with `${`, followed by the name of the variable, and end with `}`. An example of a variable reference that refers to a `java.home` system property is:
+
+```
+${java.home}
+```
+
+The system properties can be specified by
```
michaelranaldo: "System properties can be specified by:"

documentation/core-documentation/configuration/configuration.md
OndrejM: Where? I don't understand...

documentation/core-documentation/configuration/var-substitution.md
```diff
@@ -0,0 +1,45 @@
+# Variables in the configuration
+
+Variable references in the configuration are useful for reusing the same configuration in several scenarios, e.g. for several instances in the same cluster, or to provide an environment-specific configuration from the environment later when Payara server is started.
```
mikecroft: Capital "S" in Payara Server

documentation/core-documentation/configuration/var-substitution.md
mikecroft: I would prefer something like "System properties can be referenced by variables which start with `${`...etc" since "specify" also suggests that it is how they are set.

documentation/core-documentation/configuration/configuration.md
mikecroft: I think he means the definite article here - I don't see that there's a need for a convention of "Documentation" over "The documentation" or vice versa. I doubt many people would notice.

documentation/core-documentation/configuration/var-substitution.md
OndrejM: @mikecroft, this sentence really describes how they are set, not referenced - a value for a system property can be set by one of the options in the bullet list. Once set, it can be referenced. How to reference then is described in the paragraph above, at the beginning of the section about system properties. Or am I missing something? I should probably clarify that "the values of system properties" can be specified, instead of just "system properties" can be specified. Would that be understandable then?

documentation/core-documentation/configuration/configuration.md
michaelranaldo: I did say nitpicky...

documentation/core-documentation/configuration/var-substitution.md
michaelranaldo: In my case it was a definite article + @Pandrex247 said he wanted to see a colon

OndrejM: I agree, @michaelranaldo, I'll change it. My comment was an answer to Mike's comment. 

If comments are not directly related, It is better to create a separate comment, even if for the exact same line (it's possible to click on the plus sign again, even the line already contains a comment ;-) ).  We can then discuss the comments in separate threads.

documentation/core-documentation/configuration/configuration.md
OndrejM: I see, it is much clearer to just write "remove `The` before documentation", optionally followed by explanation :)
I'll delete it.

documentation/core-documentation/configuration/var-substitution.md
OndrejM: I'll change it it:

> Values of system properties can be specified by:

documentation/core-documentation/configuration/var-substitution.md
OndrejM: OK. I'll add "using **other system properties or** environment variables`, because that's also true:

>When custom properties are specified, their value can also contain other variable references. It is therefore possible to define a system property using other system properties or environment variables.

PR #125, opened 2017-02-15
Payara Server Public API Documentation [PYSVC - 248]

fturizo:
Implemented article to document the new Public API.

_This PR is a recreation since the older one got accidentally deleted._


documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+* [4. Current Interfaces](#4-current-interfaces)
+* [5. Github Repository](#5-github-repository)
+
+# 1. Overview
+
+Starting from release _4.1.1.171_, a new artifact has been created to gather all interfaces that compose the Public API for both Payara Server and Payara Micro.
+
+# 2. Use Cases
+
+This Public API should be used on the following scenarios:
```
OndrejM: > should be used **in** the following

documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+
+1. Java EE Developers and end users that develop web and enterprise applications that require Payara specific features outside the specifications \(and JCache\). 
+2. Contributors that want to develop extensions to Payara Server or Payara Micro.
+
+Previous to release _4.1.1.171_, if developers wanted to access these interfaces they had to declare a dependency to the `payara-micro` or `payara-embedded-all` artifacts. Since those dependencies are heavier and would include a lot of irrelevant classes and interfaces, the new API was created to simplify this scenario.
```
OndrejM: > dependency **on** the

documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+
+# 3. Maven Coordinates
+
+To access the public API, using **Maven** is the recommended way to go. Set the following dependency in your project's POM:
```
OndrejM: I'd rewrite to:

> The public API is released as a maven artifact in Maven Central. To use it in a maven project, set the following dependency in your project's POM:

documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+    <groupId>fish.payara</groupId>
+    <artifactId>payara-api</artifactId>
+    <version>{VERSION}</version>
+    <scope>provided</scope>
+</dependency>
+```
+
+Where the `{VERSION}` of the artifact corresponds with the version of Payara Server \(starting from _4.1.1.171_\)
+
+This artifact is published on the **Maven Central** repository, so no additional repository definitions are needed.
```
OndrejM: I think this sentence is not necessary, especially if you rewrite the sentence above as I suggested.

documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+
+# 4. Current Interfaces
```
OndrejM: I think this section is not necessary - it is hard to keep it up to date. In fact, even for 171, there is more API provided in the artifact than documented here (CDI bus, generated cache key at least). It should be enough to check the contents of the maven artifact. We should rather publish javadoc for the artifact latet and link to it from the docs.

documentation/extended-documentation/app-deployment/public-api.md
```diff
@@ -0,0 +1,51 @@
...
+
+Currently, the following interfaces are included in the API:
+
+* `fish.payara.cdi.jsr107.imp.NamedCache` \(Used for **JCache** propietary features\)
+* `fish.payara.requesttracing.api.Traced` \(Used for **Request Tracing**\)
+
+In the future, more interfaces used for Payara Server extensions will be moved to this API.
+
+# 5. Github Repository
```
OndrejM: I'd remove this section too. For users, it's not relevant where the source is available, and it is subject to change. Steve already suggested that it would be better to keep the source of the API artifact with the other sources in the main Payara repo.

OndrejM:
The page https://payara.gitbooks.io/payara-server/content/v/next/documentation/payara-micro/maven/maven.html should be also updated to reflect the need to depend on the Payara-API artifact and JCache-API artifact, and possibly Hazelcast aritfact, because those are not provided by the payara-micro.jar at build time, only at run time.

documentation/extended-documentation/app-deployment/public-api.md
fturizo: I agree, however in the current release the API lives in a separate repository so it should be documented for contributors' sake. On future releases we might remove the section and clarify this if needed.

documentation/extended-documentation/app-deployment/public-api.md
fturizo: Ok

documentation/extended-documentation/app-deployment/public-api.md
fturizo: Ok

documentation/extended-documentation/app-deployment/public-api.md
fturizo: Sounds good, I'll change it.

documentation/extended-documentation/app-deployment/public-api.md
fturizo: Agreed.

documentation/extended-documentation/app-deployment/public-api.md
fturizo: Mmm, I agree. I wrote it tentatively to gather opinions. I'll remove it.

fturizo:
Makes sense to add this information to that page. I'll see what it can be added to this feature.

PR #128, opened 2017-02-15
PYSVC-273 Added log-notifier.md page with instructions on using get-l…

michaelranaldo:
…og-notifier-configuration


documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -0,0 +1,14 @@
...
+## Configuration
+
+Payara Server by default will send notifications from the [Notification service](/documentation/extended-documentation/notification-service/notification-service.md) to the logs when activated.
+
+## Log Configuration
+
+To get the current configuration of the log notifier, run the command below:
+
+```Shell
+asadmin get-log-notifier-configuration
```
OndrejM: No set command?

michaelranaldo: Ah wasn't on the JIRA, I'll pop it in

mikecroft:
Remember to change the `SUMMARY.md` when adding new pages

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -0,0 +1,24 @@
+# Log Notifier
+The default output for the notification service is the `server.log`. More details on Payara Logging can be found [here](../core-documentation/logging/logging.md).
```
mikecroft: This link is broken. Use absolute links, not relative. Also, try and avoid "click here" style links. Try to use links that describe, or at least name, the destination.

mikecroft: This could be described better, actually - the log notifier is the only notifier that is enabled by default in the notification service. It sends to the server.log because that's how the *log notifier* works, not because that's how the *notification service* works. i.e. the service knows nothing about the server.log, it just hands off notifications to whatever notifier is enabled.

mikecroft: It would also be more accurate to say that the log notifier would log notifications to the configured instance log file, which is either the server.log file or cluster.log file depending on the configuration of the instance.

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Configuration
+
+By default, Payara Server will send notifications from the [Notification service](/documentation/extended-documentation/notification-service/notification-service.md) to the logs when activated.
```
mikecroft: Again, this sentence could be clearer. The thing that is doing the work is the log notifier, not just  "Payara Server". It also seems like the notification service is passive in this scenario, like a queue, which isn't correct. The notification service receives data from other services and sends notifications to enabled notifiers.

mikecroft: This sentence actually would be better in the introduction - it's not really about configuration

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -0,0 +1,24 @@
...
+
+### Payara Configuration
+
+The log notifier is not automatically enabled when Payara starts - to enable log notifications run the command:
```
mikecroft: This is very confusing - is it enabled by default or not? We need to describe the behaviour here accurately and clearly. Make sure you consider how this is done in the admin console too, because I'm sure this *is* enabled by default under certain circumstances

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -0,0 +1,24 @@
...
+
+```Shell
+asadmin notification-log-configure --enabled=true --dynamic=true
+```
+
+This will also enable the notification service. More details on specific commands for the notification service can be found [here](/documentation/extended-documentation/notification-service.md).
```
mikecroft: "...can be found on the Notification service page"

mikecroft: > This will also enable the notification service.

This needs to be more than an offhand remark. Under what circumstances? Is this always the behaviour? Why?

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -1,20 +1,20 @@
...
 
 ## Configuration
 
-By default, Payara Server will send notifications from the [Notification service](/documentation/extended-documentation/notification-service/notification-service.md) to the logs when activated.
+By default, the log notifier will receive notifications from the [notification service](/documentation/extended-documentation/notification-service/notification-service.md) and outputs them to the configured logs when activated.
 
 ### Payara Configuration
 
-The log notifier is not automatically enabled when Payara starts - to enable log notifications run the command:
+The log notifier is, by default, disabled on startup. However, as the default notifier it is enabled when the [notification service](/documentation/extended-documentation/notification-service/notification-service.md) is enabled.
```
mikecroft: > The log notifier is, by default, disabled on startup.

Disabled on startup doesn't sound right - surely you don't have to enable it each time you restart the server?

> However, as the default notifier it is enabled when the notification service is enabled.

This isn't very clear. This behaviour is already very confusing so we need to be both *comprehensive* and *clear*. It's not obvious what causes what and which is the preferred method for enabling or disabling things at the moment.

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -1,20 +1,20 @@
...
+
+To enable the log notifier specifically, run the command:
 
 ```Shell
 asadmin notification-log-configure --enabled=true --dynamic=true
 ```
 
-This will also enable the notification service. More details on specific commands for the notification service can be found [here](/documentation/extended-documentation/notification-service.md).
```
mikecroft: This extra sentence was quite useful.

documentation/extended-documentation/notification-service/notifiers/log-notifier.md
```diff
@@ -1,20 +1,20 @@
 # Log Notifier
-The default output for the notification service is the `server.log`. More details on Payara Logging can be found [here](../core-documentation/logging/logging.md).
+The default output for the log notifier is the configured instance log file, which is either the `server.log` file or `cluster.log` file, depending on the instance configuration. More details on Payara Logging can be found [on the logging page](/documentation/core-documentation/logging/logging.md). The log notifier is the only notifier which is enabled by default when the notification service is activated.
```
mikecroft: This is an odd introduction. It reads like I'm already halfway through the page - maybe the final sentence should be the first?

PR #134, opened 2017-02-20
PYSVC-245 PYSVC-244 New page on new JAR structure

michaelranaldo:


documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,26 @@
...
+
+To make Payara Micro easier to work with and increase its future extensibility, Payara Micro now uses a nested JAR format using code from Springboot. This has changed both the classloader and its behaviour. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
+
+#### Default Packaging
+
+By default, Payara Micro will unpack the nested JARs into folders within the file system and then load them as classes. This is  faster to boot than previous versions of Payara Micro. This requires no explicit command-line arguments.
+
+#### `--nested`
+
+The other option now available is launching Payara Micro with the `--nested` flag. This will load the classes directly from the nested JARs to the memory. While this slows boot by approximately 13%, the nested flag optimizes the nested JARs for random access.
```
mikecroft: >The other option now available is launching Payara Micro with the `--nested` flag.

Other than what? This is the only option you've mentioned. Try to stick to either saying "option" or "flag", rather than mixing them.

mikecroft: How does using the `--nested` option optimise the JARs? Isn't this a bit redundant since there is no way for them *not* to be "optimised" and the slower boot time would include that optimisation anyway?

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,26 @@
...
+
+To load Payara Micro with the `--nested` flag:
```
mikecroft: > To load Payara Micro with the `--nested` flag:

To start Payara Micro as a nested JAR, use the `--nested` option as shown:

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,26 @@
...
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
+There are a number of ways to structure an Uber-JAR - a jar file which contains both a project and all its dependencies. Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR. The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
+
+This made Payara Micro awkward to work with, and lead to dependency issues.
+
+### Nested JAR
+
+To make Payara Micro easier to work with and increase its future extensibility, Payara Micro now uses a nested JAR format using code from Springboot. This has changed both the classloader and its behaviour. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
```
mikecroft: Spring Boot is two words

mikecroft: > To make Payara Micro easier to work with and increase its future extensibility

For whom? This change is internal and doesn't really affect end-users

mikecroft: > This has changed both the classloader and its behaviour.

Has the behaviour changed in a way that is noticeable to end-users at all, or relevant to the documentation?

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,26 @@
+# Payara Micro Jar Structure
+
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
+There are a number of ways to structure an Uber-JAR - a jar file which contains both a project and all its dependencies. Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR. The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
```
michaelranaldo: Would "Payara Micro's Uber-JAR structure" suffice or should I use a different term?

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,25 @@
...
+Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR.
+The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
+
+### Nested JAR
+
+Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
+
+#### Default - Unpacking to File System (`--unpack`)
+
+By default, Payara Micro will unpack the nested JARs into a temporary directory within the directory specific by the system property `java.io.tmpdir` and then load them as classes. This is  faster to boot than previous versions of Payara Micro. This requires no explicit command-line arguments.
```
OndrejM: > This is faster to boot

I don't think it is faster than the previous way of packaging. It's probably a bit slower, but we may say it's comparably fast as the previous way of packaging, while adding many advantages.

OndrejM: The advantages would be isolation of modules that prevents class collisions, safer way to extend the basic JAR package by additional modules, which we plan to support in the future, and maybe more that I don't remember. 
There is also a disadvantage that , at compile time, the JAR doesn't provide anything but a few core classes. It is necessary to use the new Payara-API artifact to build applications that rely on Payara-Micro specific API. Also, there are problems with embedding Payara Micro into an app (not sure if even possible), what was previously a piece of cake.

We don't have to mention this in the docs, I just wanted to note it :) You may pick what you think is important and add it into docs, though.

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,28 @@
...
+The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
+
+### Nested JAR
+
+Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
+
+#### Default - Unpacking to File System (`--unpack`)
+
+By default, Payara Micro will unpack the nested JARs into a temporary directory within the directory specific by the system property `java.io.tmpdir` and then load them as classes.
+This results in a boot time comparably to the original boot, but with the advantages of fewer class collisions and safer extensibility.
```
mikecroft: > This results in a boot time comparably to the original boot

doesn't make sense. What is "original boot" in this context?

mikecroft: > fewer class collisions and safer extensibility.

Is this accurate? It sounds like you're saying `--nested` is unsafe to use.

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Uber-JAR Structure
+Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR.
+The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
+
+### Nested JAR
+
+Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
+
+#### Default - Unpacking to File System (`--unpack`)
```
mikecroft: Starting the title with "Default" is a bit awkward. Default isn't the title. It should be at the end, or left out altogether

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,28 @@
+# Payara Micro Jar Structure
+
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
+Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR.
+The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
```
mikecroft: *Payara Micro

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,28 @@
+# Payara Micro Jar Structure
+
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
+Previously, Payara Micro inherited the shaded jar format of Payara Embedded, where dependencies which shared packages were _shaded_ and renamed if they were not unique and all JARs were unpacked, then packed within a single JAR.
```
mikecroft: This sentence is very long and rambling with lots of commas and multiple `and`s. Also, are you sure about the Payara Embedded part?

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -1,19 +1,20 @@
...
 
-Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
+Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara Micro classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
 
-#### Default - Unpacking to File System (`--unpack`)
+#### Unpacking to File System (`--unpack`)
 
 By default, Payara Micro will unpack the nested JARs into a temporary directory within the directory specific by the system property `java.io.tmpdir` and then load them as classes.
-This results in a boot time comparably to the original boot, but with the advantages of fewer class collisions and safer extensibility.
+This change to Payara Micro results in a comparable boot time to previous Payara Micro releases, but with the added advantages of fewer class collisions and safer extensibility.
```
OndrejM: I would reword "This change to Payara Micro results..." into "This way of booting Payara Micro results..."

OndrejM: Also this "..added advantages of fewer class collisions..." to "added advantages of avoiding class collisions"

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
+# Payara Micro JAR Structure
+
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
+Before 4.1.1.171, Payara Micro used a shaded JAR format.
+Shared packages were renamed and all of Payara Micro's dependencies were unpacked, then packed within a single JAR.
```
smillidge: packages weren't renamed. I'm not sure it is really a shaded format as it was just unpacked and recombined.

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+The JVM classloader handled all of the internal classes and resources, with the Payara classloader only handling WAR files.
```
smillidge: Hmm this statement is likely to be not correct or very simplistic and not really relevant for the user.

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Nested JAR
+
+Payara Micro now uses a nested JAR format using code from Spring Boot. The Payara Micro classloader now handles the majority of internal classes and resources (leaving only a few essential classes to the JVM classloader) and  there are now two options for unpacking classes:
```
smillidge: The nested classloader is from SpringBoot but by default it is not used so no need to mention this here.

smillidge: The classloader comments are a bit weird. What is the Payara Micro classloader? If we go into classloading we need a deeper explanation which is accurate and describes the changed class loading correctly.

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+
+#### Unpacking to File System (`--unpack`)
+
+By default, Payara Micro will unpack the nested JARs into a temporary directory within the directory specific by the system property `java.io.tmpdir` and then load them as classes.
```
smillidge: specific should be specified. However this can also be changed using --unpackdir

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+This change to Payara Micro results in a comparable boot time to previous Payara Micro releases, but with the added advantages of fewer class collisions and safer extensibility.
```
smillidge: not sure what fewer class collisions means

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+This requires no explicit command-line arguments.
+
+#### Unpacking to Memory (`--nested`)
+
+The other flag now available is `--nested`.
+This will load the classes directly from the nested JARs to the memory, but may slow boot.
```
smillidge: classes are always loaded to memory so not sure what this statement means

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
+# Payara Micro JAR Structure
+
```
smillidge: Some description of what the jar structure is and what each folder in the jar does would be useful otherwise this doc is all information a user doesn't care about.

documentation/payara-micro/payara-micro-jar-structure.md
OndrejM: @michaelranaldo probably meant that there is not additional step of unpacking the classes before they are loaded. I would add:

> ...to the memory, *without unpacking the JARs into a folder*, but may slow boot.

documentation/payara-micro/payara-micro-jar-structure.md
michaelranaldo: Would it be better to have the focus of this doc to be more on navigating and using the actual jar then rather than the behind the scenes?

OndrejM: What's written already is very useful, but I would add a section about the structure of files inside the JAR (before the "Nested JAR" section).

documentation/payara-micro/payara-micro-jar-structure.md
```diff
@@ -0,0 +1,29 @@
...
+From Payara 4.1.1.171, Payara Micro has changed its JAR structure.
+
+## Uber-JAR Structure
```
OndrejM: This structure doesn't apply only to the Uber-JAR, but also to payara-micro.jar. Rename the section or just remove the title and the sentence above it, because it is redundant - the text below can be placed directly under the main page title, without any sub-titles.

PR #135, opened 2017-02-20
PYSVC-270 Added in new (empty) SNMP notifier page and get-notifier-co…

michaelranaldo:
…nfig


mikecroft:
Fabio is working on the SNMP notifier. We can't have PRs which do half a change or act as placeholders.

PR #140, opened 2017-02-20
PYSVC-219 Added detail on using historicaltraceenabled to healthcheck

michaelranaldo:


mikecroft:
Looks good @michaelranaldo but be careful with introducing formatting changes that don't affect anything. There are lots of spaces deleted here that shouldn't really be in a commit like this.

PR #145, opened 2017-02-22
Bringing "Next" into line with master

mikecroft:


mikecroft:
Some of the above are basically empty commits and mostly there to just bring the histories of both branches together since they'd diverged a bit.

PR #150, opened 2017-03-02
PYSVC-186 Added CTRL-C as a way to shut down Micro Instances

michaelranaldo:
Since 171

### DIFF
--- a/documentation/payara-micro/stopping/stopping.md
+++ b/documentation/payara-micro/stopping/stopping.md
@@ -1,15 +1,15 @@
 # Stopping an Instance
 This section describes how to shut down a Payara Micro instance.
 
-## 6.1 Stopping an Instance from the Command Line
-_Since Payara 4.1.1.171_
+## Stopping an Instance from the Command Line
 
 Payara Micro instances can be stopped by either:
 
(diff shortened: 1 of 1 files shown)


documentation/payara-micro/stopping/stopping.md
```diff
@@ -2,6 +2,8 @@
 This section describes how to shut down a Payara Micro instance.
 
 ## 6.1 Stopping an Instance from the Command Line
+_Since Payara 4.1.1.171_
```
OndrejM: It would be better to synchronize on the same convention for "Since".
I already used this form:
https://github.com/payara/Payara-Server-Documentation/blob/master/documentation/extended-documentation/cdi-events.md

Not saying it's better, but it could look the same :)

mikecroft: Good point - if we can get the move to asciidoc done soon, this should be something to look at. We could have a standard "note" format or something. Consistency is good!

michaelranaldo: This would be great with ASCIIdoc with the footer: tag

PR #155, opened 2017-03-03
PYSVC-308 Moved set-request-tracing to top as it is the most wholesom…

michaelranaldo:
…e command


OndrejM:
I added this commit to fix the missing title of the first section: https://github.com/payara/Payara-Server-Documentation/commit/b17864a24c0cd103db2e4d984e7724815af1c5cf

PR #156, opened 2017-03-03
PYSVC-256 Added paragraph to advanced-jdbc-configuration-and-diagnost…

michaelranaldo:
…ics explaining Validation changes.

Unsure if this is the best placement - would core-doc be more suitable?

### DIFF
--- a/documentation/extended-documentation/advanced-jdbc/advanced-jdbc-configuration-and-diagnostics.md
+++ b/documentation/extended-documentation/advanced-jdbc/advanced-jdbc-configuration-and-diagnostics.md
@@ -1,4 +1,4 @@
-#Advanced JDBC Configuration and Diagnostics.
+# Advanced JDBC Configuration and Diagnostics.
 _Payara Server and Micro 161 (4.1.1.161) onwards_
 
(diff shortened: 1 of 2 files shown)


documentation/extended-documentation/advanced-jdbc/advanced-jdbc-configuration-and-diagnostics.md
```diff
@@ -26,6 +26,12 @@ All connection pool properties available on the administration console for Payar
 
 See [Advanced Connection Pool Properties](advanced-connection-pool-properties.md) for detailed information.
 
+## JDBC Connection Validation Changes
+
+From Payara 4.1.1.171 the GlassFish API has had a slight change to allow for timeouts with the `table` and `custom-validation` connection validation types.
```
OndrejM: I think these implementation details are not relevant, as nobody can access the modified API (there's no API artifact that provides it). We don't want to document the API change. The relevant thing to document is that now, `table` and `custom-validation` validators take _Statement Timeout_ (configured for the connection pool) into account (by passing it to the `isConnectionValid` method). Before, there was no timeout when performing the validation, which was very unfortunate.

OndrejM:
If you move the part about statement timeout to the core docs, don't forget to add a link between the core part and this part in extended docs (both ways if makes sense).

PR #164, opened 2017-03-22
PYSVC 246 - Preboot and Postboot command scripts for Payara Micro 

fturizo:
Restructured the articles on running administration commands for Payara Micro instances and added a new article documenting the usage of _preboot_ and _postboot_ command scripts.


fturizo:
Conflicts resolved, @mikecroft. Let me know of any fixes to make to the articles.

OndrejM:
> @mikecroft: I'm also not sure about the SUMMARY.md changes

I think that Github diff just messes up - in a decent diff (netbeans), I only see the changed lines. Also in github, you can see which lines changed, but you have to switch to the rich diff (the icon to the left of the View button). For some reason, github thinks that several lines were changed (maybe EOL characters), but a real change is only on 3 lines.

mikecroft:
Yeh, I think it must be EOL characters. I can't think of any other reason!

PR #166, opened 2017-03-24
PYSVC 199 - SNMP Notifier Documentation

fturizo:
Documentation finished. Waiting to merge until the feature is fixed.


OndrejM:
In the summary file, only SNMP Notifier link was added.
I will now revert the changes to SUMMARY file, so that the page is not visible, and will create a PR to add this link later, once the feature is ready.

mulderbaba:
@OndrejM what's the problem on this one? do we have an internal task about it?

OndrejM:
I think we only discussed in the chat - Fabio couldn't observe any notifications on SNMP, probably because he is using a different SNMP tool than you were using. Therefore he was not confident that it works. Please connect with @fturizo for details. I think no issue is created for this.

PR #173, opened 2017-04-07
fixes for general info section + Payara Micro section

mikecroft:


OndrejM:
Isn't it too much work to fix all the line breaks? Maybe it's less work to compare and merge the pages generated in my branch.

mikecroft:
fixing line breaks takes no time at all, I've got a macro to fix all with one command. It makes sense to do it as I go, plus it gives me a nice way to tick off the paragraphs I've checked :)

mikecroft:
I don't think it's necessary, though, those breaks don't seem to be preserved in Gitbook

mikecroft:
I agree @OndrejM. There may still be more work to do, but any work would build on these changes, since I only corrected formatting. I'll merge now.

PR #201, opened 2017-05-24
added to app deployment stuff about pre/post deployment

mikecroft:


mikecroft:
Is this enough? Any more needed?

documentation/extended-documentation/app-deployment/app-deployment.adoc
```diff
@@ -1,11 +1,46 @@
...
++
+*Aim*: passing a list of commands as a file to the server to run post-boot.
+This is where commands such as service configuration commands should be run.
+
+[NOTE]
+====
+There are some important caveats to note:
+
+* Both files should be saved in a text format, with each command on it's own line.
+* Lines starting with a '#' are treated as comments
```
OndrejM: A line is also ignored when there is a blank space before #, and also empty and blank lines are ignored.
Lines with an erroneous command are also ignored, with an Exception in the logs.

mikecroft:
@OndrejM I created an internal ticket for the help as well :smile: it might even be number 1708! It's definitely an important one, though, since there are one or two other places the help needs an improvement or fix

documentation/extended-documentation/app-deployment/app-deployment.adoc
mikecroft: Would you reword that line to be:

> Lines where the first non-whitespace character is `#` are treated as comments

OndrejM: Makes sense. But I would add another sentence:
"The files can include blank lines which will be ignored."

mikecroft:
Added a couple of extra lines @OndrejM

PR #202, opened 2017-05-25
Changed XMPP references to room name to room id

MattGill98:
The get-xmpp-notifier-configuration command still has room name as a column heading which will need changing. I'll make a new PR for it under the original JIRA


mikecroft:
Well spotted! Cheers

PR #229, opened 2017-08-04
Micoprofile config

mikecroft:


documentation/microprofile/README.adoc
```diff
@@ -0,0 +1,71 @@
...
+|
+* JAX-RS 2.0
+* CDI 1.1
+* JSON-P 1.0
+* http://microprofile.io/project/eclipse/microprofile-config[Config 1.0]
+
+| [source, xml]
+----
+<dependency>
+    <groupId>org.eclipse.microprofile</groupId>
```
OndrejM: MP 1.1 hasn't been released and this is still subject to change. Personally, I would change artifactId and type if possible and everybody agrees.

documentation/microprofile/config.adoc
```diff
@@ -0,0 +1,302 @@
...
+== Background
+The Config API was created to allow for the modification of application
+configuration from outside the application without needing the application to be
+repackaged.
+
+A full overview for the reasoning behind the API can be found in
+https://github.com/eclipse/microprofile-config/blob/master/README.adoc[the
+README for the source code repository.]
+
+https://github.com/eclipse/microprofile-config/blob/master/spec/src/main/asciidoc/microprofile-config-spec.asciidoc[
```
OndrejM: I would rather link to the Config 1.0 release page: https://github.com/eclipse/microprofile-config/releases/tag/1.0
It contains Javadoc, PDF spec and other info.

documentation/microprofile/config.adoc
```diff
@@ -0,0 +1,302 @@
...
+|Source
+|Ordinal
+|Description
+
+|`domain`
+|110
+|Retrieves and stores config values in the domain.xml as key values. These will
+then apply to all instances within a domain
+
+|`config`
```
OndrejM: To be honest, from this description, I have no idea how this works

OndrejM: Part of it is that the sentence doesn't make sense. Should it be 2 separate sentences?

documentation/microprofile/config.adoc
```diff
@@ -0,0 +1,302 @@
...
+configuration
+
+|`server`
+|130
+|Retrieves and stores config values at the level of the server in the domain.xml.
+Property values will only apply on the named server instance
+
+|`application`
+|140
+|retrieves and stores config values at the level of the application. Property
```
OndrejM: Captial R in **R**etrieves and stores

documentation/microprofile/config.adoc
```diff
@@ -0,0 +1,302 @@
...
+
+== Converters
+In addition to the standard converters specified by the API, Payara's implementation
+includes out-of-the-box converters for URL and InetAddress. These have `@Priority`
+of `1`, the same as the default converters.
+
+
+== Asadmin Commands
+
+IMPORTANT: A `--source` value of `jndi` is not supported for this command in
```
OndrejM: I would move this notice under the *Usage* section of the set-config-ordinal` command. In this place it is rather confusing to which command it relates to.

documentation/microprofile/README.adoc
OndrejM: I mean, we shouldn't change this now, but update it once MP 1.1 artifact is finally released.

documentation/microprofile/config.adoc
mikecroft: It relates to all commands. It was originally there, but when I noticed it was missing from everywhere, I moved it. Should be reworded.

OndrejM: As it is now, it looks pretty scary - before reading any useful documentation, the reader hits an IMPORTANT warning...

I would rather move it at the end of the page, if it applies to all commands and also reword it to make sense. It;s not a big deal because it should be fixed in the next version, but the psychological impact could be big if it stays like this.

documentation/microprofile/README.adoc
mikecroft: Agreed

documentation/microprofile/README.adoc
```diff
@@ -0,0 +1,71 @@
...
+* JAX-RS 2.0
+* CDI 1.1
+* JSON-P 1.0
+
+| [source, xml]
+----
+<dependency>
+    <groupId>io.microprofile</groupId>
+    <artifactId>microprofile</artifactId>
+    <version>1.0.0</version>
```
mikecroft: changed

documentation/microprofile/config.adoc
mikecroft: Clarified

PR #244, opened 2017-09-04
Pysvc 342 document historic request tracing

MattGill98:
- Added list-historic-requesttraces command to list of request tracing commands.
- Updated relevant pages in request tracing folder to match new configuration options.


mikecroft:
Looks good - might be better to raise a PR against the 174 branch, and then raise a PR from 174 to Master so the commit hashes all stay the same. @OndrejM WDYT?

OndrejM:
After we changed the release procedure, this can be merged into the master branch as is.

PR #245, opened 2017-09-05
[PAYARA-1751] - Documented Adding Libraries in Payara Micro

fturizo:
Documented the new feature introduced in 173 for adding 3rd party libraries to a Payara Micro instance.


documentation/payara-micro/adding-jars.adoc
```diff
@@ -0,0 +1,108 @@
...
+
+Here is an example to package 2 JDBC drivers (for MySQL and PostgreSQL) into an
+Uber JAR called `my-micro.jar`:
+
+[source, shell]
+----
+java -jar payara-micro-4.1.2.173.jar --addJars postgresql-9.4.1208.jre7.jar:mysql-connector-java-5.1.44-bin.jar --outputUberJar my-micro.jar
+----
+
+Which we can inspect to verify that the JAR files were packaged successfully:
```
OndrejM: The word "Which" in the beginning doesn't make sense to me. We can omit it: "We can inspect [uber JAR contents] to verify that..."

fturizo: Makes sense, changes implemented.

PR #247, opened 2017-09-06
Basic admonition icons

OndrejM:
Adds a plugin that turns admonition captions like WARNING or NOTE into icons, like :warning: and :information_source: 

Preview: https://ondrejm.gitbooks.io/payara-server-documentation-fork/content/v/admonition-icons/documentation/payara-micro/adding-jars.html

### DIFF
--- a/book.json
+++ b/book.json
@@ -1,6 +1,7 @@
 {
     "plugins": [
         "expandable-chapters",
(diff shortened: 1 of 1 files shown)


mikecroft:
Wonderful! I thought we'd be stuck doing this with more custom CSS but this is great!

PR #252, opened 2017-09-12
PYSVC-517 Mbean docs

Cousjava:


OndrejM:
Looks good, although it's just too long to read it all.
The page mbeans.adoc is very long, might be worth splitting it to many smaller pages.

documentation/user-guides/monitoring/monitoring.adoc
```diff
@@ -20,7 +20,7 @@ new guides get added, this section will include:
 link:enable-jmx-monitoring.adoc[How do I enable JMX monitoring in Payara Server?]::
 Before anything else, we need to know how to enable JMX monitoring and get
 familiarised with the concepts involved.
-What MBean metrics should I monitor?::
+link:mbeans.adoc[What MBean metrics should I monitor?]::
```
OndrejM: I don't see a link to mbeans.adoc in the SUMMARY.adoc - it needs to be there to generate the page.

mikecroft:
Yes, I agree it is longer than I'd expected. I think this is probably the clearest way to represent the information, though - I don't know how it could be done better.

fturizo:
@OndrejM, are the requested changes good? Can you approve the PR if so?

mikecroft:
It looks like there's a problem with this PR - the page doesn't exist. I think it's just that the link in the SUMMARY.adoc is wrong. It's in the wrong place and links to a doc that doesn't exist. It should be in the User Guides section, not the JMX logging service @Cousjava can you create a new PR to correct the Summary link please?

PR #262, opened 2017-10-10
PAYARA-2070 Document New JMX Monitoring Notification Commands

michaelranaldo:


mikecroft:
Looks good at first glance @michaelranaldo! Remember to add the new page to the SUMMARY.adoc though

michaelranaldo:
Gah! Added it...

documentation/extended-documentation/jmx-monitoring-service/asadmin-commands.adoc
```diff
@@ -20,29 +20,72 @@ reload the configuration.
...
+|Type
+|Description
+|Default
+|Mandatory
+
+|`--target`
+|String
+|The instance or cluster to get the monitoring
+service configuration of
+|`server-config`
```
OndrejM: If this is true, then the description should change to something like:
"The name of the configuration to get the monitoring service configuration from"

documentation/extended-documentation/jmx-monitoring-service/asadmin-commands.adoc
```diff
@@ -57,10 +100,420 @@ Retrieves the monitoring service configuration and outputs it.
...
+|Type
+|Description
+|Default
+|Mandatory
+
+|`--target`
+|String
+|The instance or cluster to get the monitoring
+service configuration of
+|`server`
```
OndrejM: The default for `set-monitoring-configuration` is updated to `server-config` instead of `server`. Should this be updated also here?

documentation/extended-documentation/jmx-monitoring-service/asadmin-commands.adoc
OndrejM: However, can you make sure that `server-config` is the default and not `server`? What about other asadmin commands on this page? They also have `--target` argument and this page says that default value is `server` - it's strange that only this single asadmin command has the default value `server-config`.

documentation/extended-documentation/jmx-monitoring-service/jmx-notification-configuration.adoc
```diff
@@ -0,0 +1,145 @@
+= JMX Notification Configuration
+
+JMX Monitoring can be configured to send notifications using the
+link:/documentation/extended-documentation/notification-service/notificationservice.adoc[notification service].
```
OndrejM: Fix the link - it should be `notification-service.adoc`

documentation/extended-documentation/jmx-monitoring-service/jmx-notification-configuration.adoc
```diff
@@ -0,0 +1,145 @@
...
+== Adding Notifiers via asadmin
+
+For each notifier you wish to enable use the command:
+
+[source, shell]
+----
+monitoring-<notifier>-notifier-configure --enabled true
+----
+
+Available Notifiers:
```
OndrejM: I would rather link to the [Notifiers page](https://github.com/payara/Payara-Server-Documentation/blob/master/documentation/extended-documentation/notification-service/notifiers.adoc) instead of maintaining the list also here - it could get out of synch easily if we add more notifiers.

documentation/extended-documentation/jmx-monitoring-service/jmx-notification-configuration.adoc
```diff
@@ -0,0 +1,145 @@
...
+
+You can now add your monitoring attributes to the notification service using
+the following values:
+
+* _name_: the MBean attribute name
+* _value_: the MBean domain name
+* _description_: displayed in the `get-monitoring-configuration` asadmin
+command, _optional_
+
+For example, to monitor the `activeworkcount` of the MBean on the DAS with a
```
OndrejM: I would also explain shortly how to get the string of the MBean domain using e.g. JConsole - a screenshot of JConsole showing the `ObjectName` property of the `activeworkcount`  attribute should be enough.

PR #264, opened 2017-10-16
PSYVC-508 removed documented removed commands

Cousjava:


fturizo:
@Cousjava, why was the documentation from these commands removed? I recall that some of these commands were redundant, but only that.

Cousjava:
@fturizo  Because the commands themselves have been removed from Payara in 174.

PR #268, opened 2017-11-17
Add 174

Pandrex247:


OndrejM:
I needed to fix some commas so that the Gitbook docs version selector works. It helps to check the JSON with https://jsonchecker.com/.

PR #270, opened 2017-11-22
Added note for machine meory usage only working on Linuz

Cousjava:


documentation/extended-documentation/health-check-service/README.adoc
```diff
@@ -21,6 +21,8 @@ to one or more notifiers, e.g. a log file.
 This allows operations teams to rapidly detect problems or work out what happened after these
 problems have occurred.
 
+Note that the Host Memory Usage currently only works on Linux and *BSD.
```
OndrejM: Change this please to the following so that it displays as INFO admonition as documented [here](http://asciidoctor.org/docs/user-manual/#admonition):

NOTE: The Host Memory Usage currently only works on Linux and BSD derivatives.

Cousjava: done

PR #287, opened 2018-01-24
PYSVC-91 Document the Hazelcast EJB Timer Store

michaelranaldo:
Documented the EJB Timer Store. The Admin Console / Asadmin integration does not appear to be finished according to the GitHub issue so I've documented the direct domain.xml edit.

### DIFF
--- a/documentation/extended-documentation/ejb.adoc
+++ b/documentation/extended-documentation/ejb.adoc
@@ -149,15 +149,15 @@ the module, use the `--forceName` command option.
 _Since Payara Server 4.1.1.164_
 
 It is now possible to persist an EJB Timer to the Hazelcast cluster rather than
-to a database in Payara Server in the same way as [Payara Micro]. This means that the cluster itself will store
-the timer details, preserving it even if the original instance leaves the
-cluster.
+to a database in Payara Server in the same way as [Payara Micro]. This means that the 
+cluster itself will store the timer details, preserving it even if the original instance 
+leaves the cluster.
 
-CAUTION: This feature has not yet been added to the Admin Console or the
-asadamin commands. To persist an EJB Timer to the Hazelcast cluster, you will
-need to edit the domain.xml. You will need to restart the server after editing,
-and ensure that you do not modify any other sections. Create a backup copy
-beforehand.
+CAUTION: This feature is tech-preview and is subject to change. Use with caution.
+
+To persist an EJB Timer to the Hazelcast cluster, you will need to edit the domain.xml. 
+You will need to restart the server after editing, and ensure that you do not modify 
+any other sections. Create a backup copy of your domain.xml before editing.
 
 Within your config (`<config name="${your config}">`), change the section:
 
--- a/documentation/extended-documentation/ejb.adoc
+++ b/documentation/extended-documentation/ejb.adoc
@@ -148,45 +148,45 @@ the module, use the `--forceName` command option.
 
 _Since Payara Server 4.1.1.164_
 
-It is now possible to persist an EJB Timer to the Hazelcast cluster rather than
-to a database in Payara Server in the same way as [Payara Micro]. This means that the 
-cluster itself will store the timer details, preserving it even if the original instance 
+It is possible to persist an EJB Timer to the Hazelcast cluster rather than
+to a database. The same feature is available in link::/documentation/payara-micro/persistent-ejb-timers.adoc[Payara Micro].
+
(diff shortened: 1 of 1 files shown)


documentation/extended-documentation/ejb.adoc
```diff
@@ -142,3 +142,51 @@ tries to be overridden using the `-name` option.
...
+== Persistent EJB Timers with Hazelcast
+
+_Since Payara Server 4.1.1.164_
+
+It is now possible to persist an EJB Timer to the Hazelcast cluster rather than
+to a database in Payara Server in the same way as [Payara Micro]. This means that the cluster itself will store
+the timer details, preserving it even if the original instance leaves the
+cluster.
+
+CAUTION: This feature has not yet been added to the Admin Console or the
```
mikecroft: I didn't know it still doesn't even have any asadmin commands! I'm not sure I like the idea of documenting a feature where you need to edit the domain.xml, since we always say that's not supported/advise against it.

Could we perhaps create a standard "CAUTION" note format to add a disclaimer that a feature is "Tech Preview" only, which is basically what is happening here? Basically to get across the same message as this note does, but in a more general way. (CC @OndrejM @fturizo )

documentation/extended-documentation/ejb.adoc
```diff
@@ -142,3 +142,51 @@ tries to be overridden using the `-name` option.
...
 IMPORTANT: In the case you need to overwrite the name of the module when deploying
 the module, use the `--forceName` command option.
+
+[[persistent-ejb-timers-with-hazelcast]]
+== Persistent EJB Timers with Hazelcast
+
+_Since Payara Server 4.1.1.164_
+
+It is now possible to persist an EJB Timer to the Hazelcast cluster rather than
+to a database in Payara Server in the same way as [Payara Micro]. This means that the
```
OndrejM: This section shouldn't document Payara Micro. If there's a page about this feature in Payara Micro section, we could provide a link but nothing more.

I suggest rewriting to something like:

"It is possible to persist an EJB Timer to the Hazelcast cluster rather than to a database. The same feature is available in Payara Micro (link to the appropriate page in Payara Micro docs)"

If there's no section in Payara Micro docs, then please add it. It would be enough if it only contained a short description of what the feature is and a link to the documentation for Payara Server. It may also contain the caution about technical preview.

documentation/extended-documentation/ejb.adoc
```diff
@@ -142,3 +142,51 @@ tries to be overridden using the `-name` option.
...
+_Since Payara Server 4.1.1.164_
+
+It is now possible to persist an EJB Timer to the Hazelcast cluster rather than
+to a database in Payara Server in the same way as [Payara Micro]. This means that the 
+cluster itself will store the timer details, preserving it even if the original instance 
+leaves the cluster.
+
+CAUTION: This feature is tech-preview and is subject to change. Use with caution.
+
+To persist an EJB Timer to the Hazelcast cluster, you will need to edit the domain.xml.
```
OndrejM: Is there really no asadmin command for this? If there is, please document it instead of describing how it looks like in domain.xml. Even if the asadmin command would be just calling the `set` command with an appropriate dotted name for the property in domain.xml.

documentation/extended-documentation/ejb.adoc
OndrejM: If we figure out how gitbook can support included page fragents, we could then create a fragment with a sandard caution, which could be simply included in the asciidoc page. Until then, we have to agree to some common message and keep copying it.

documentation/extended-documentation/ejb.adoc
mikecroft: Good point about the `set` command @OndrejM. I've continued the discussion on the previous thread which seems to be hidden now:
https://github.com/payara/Payara-Server-Documentation/pull/287#discussion_r163606332

documentation/extended-documentation/ejb.adoc
mikecroft: Yes, I agree @OndrejM - the asciidoc include functionality would be perfect here.

michaelranaldo:
Updated with feedback. Examined using ASCIIDoc include but thought I'd leave that for a later comprehensive JIRA

OndrejM:
I'm going to merge this and continue with refactoring the docs. I have to dismiss @mikecroft 's review - I believe all raised issues have been addressed.

PR #288, opened 2018-01-26
PYSVC-75 Convert the Maven Cargo Plugin Blog to a User Guide

michaelranaldo:


documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
+[[using-the-maven-cargo-plugin]]
+= Using the Maven Cargo Plugin
+
+When creating a Java EE application, it is important that development and
+testing take place on a server that's as close to your target production
+environment as possible.
+
+By using the link::https://codehaus-cargo.github.io/cargo/Maven2+plugin.html[Maven Cargo]
```
mikecroft: Formatting error on the link

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+plugin within your Maven project, it is possible to deploy your application to
+local and remote instances of Payara Server, simplifying testing by managing
```
mikecroft: This is a bit of a long sentence. Could be worth breaking it down into 2 or 3.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+on your target environment.
+
+== Adding the Cargo Plugin
+
+The cargo plugin can be added to any Maven project by adding the following
+to the `<pluginManagement>` section of your pom:
+
+[source,xml]
+----
+<plugin>
```
mikecroft: It would be good to include the context of the `<pluginManagement>` tags as well (use an ellipsis to suggest other plugins not included)

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+ ** Runs the `undeploy` goal, undeploying the application.
+
+ - `stop`:
+ ** Runs the `stop` goal, stopping the _local_ server. This does not work for
+ remote servers.
+
+=== Deploying via the Cargo Plugin
+
+To deploy your application to Payara Server using the Maven Cargo Plugin,
+you need to add a _profile_. From a general perspective, a Maven profile serves
```
mikecroft: > ...you need to add a profile

Are you sure? This may be best practice but I don't think it's required

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+Within a Maven profile you can specify configuration for Payara Server within
+the Cargo plugin: the host, port, username, password, and importantly whether
+to use an existing or new local or remote server.
+
+To create a default local Payara Server profile, add the following within the
+`<profiles>` tags in your pom:
+
+[source, xml]
+----
+<plugin>
```
mikecroft: This definitely needs context, I assume there must be a `<profile>` tag within the `<profiles>` tag? It's not clear from this snippet.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+ ** Runs the `stop` goal, stopping the _local_ server. This does not work for
+ remote servers.
+
+=== Deploying via the Cargo Plugin
+
+To deploy your application to Payara Server using the Maven Cargo Plugin,
+you need to add a _profile_. From a general perspective, a Maven profile serves
+to collate specific configuration under a single command line flag.
+
+Within a Maven profile you can specify configuration for Payara Server within
```
mikecroft: This is another complex sentence. Try to avoid lists of lists following other lists. "existing or new local or remote server" is hard to parse.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+      <groupId>org.glassfish.main.deployment</groupId>
+      <artifactId>deployment-client</artifactId>
+      <version>4.1.1</version>
+    </dependency>
+  </dependencies>
+</plugin>
+----
+
+This is a local profile which relies on using an existing Payara Server
+installation.  You will need to specify the following within the `<properties>`
```
mikecroft: Would be worth clarifying that you don't need to do this to use the plugin, it's another best practice to use variables set as properties.

mikecroft: It's a guide, so a worked example, so you can justify telling people to do that because it's how your example works.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+</plugin>
+----
+
+This is a remote profile which relies on using an existing Payara Server
+installation.  You will need to specify the following within the `<properties>`
+tags in your pom:
+
+[source, xml]
+----
+<cargo.remote.username>admin</cargo.remote.username>
```
mikecroft: This looks wrong. The variables you use in your profile are `payara.username`.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+<cargo.remote.password>adminadmin</cargo.remote.password>
```
mikecroft: It's probably good to specify that this should be the remote password that the user has set. `adminadmin` is a very old default GlassFish password that hasn't existed in a long time. Unless the user has set this, it won't work.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+<admin.port>4848</admin.port>
+<hostname>localhost</hostname>
```
mikecroft: Again, specify here that this should be the hostname of the remote server. Your explanatory paragraph above doesn't make it clear.

SUMMARY.adoc
```diff
@@ -176,6 +176,7 @@
 *** link:documentation/user-guides/connection-pools/validation.adoc[Connection validation]
 *** link:documentation/user-guides/connection-pools/leak-detection.adoc[Statement and Connection Leak Detection]
 ** link:documentation/user-guides/bypassing-jms-connections-through-a-firewall.adoc[Bypassing JMS Connections through a Firewall]
+** link:documentation/user-guides/using-the-maven-cargo-plugin.adoc[Using the Maven Cargo Plugin]
```
OndrejM: I would rename the guide to something that mentions the goal rather than the method, e.g. "Deploying apps from Maven". If people haven't heard about the cargo plugin they wouldn't understand what the guide is about.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
+[[using-the-maven-cargo-plugin]]
+= Using the Maven Cargo Plugin
```
OndrejM: I would rename the guide as explained in the previous comment

documentation/user-guides/using-the-maven-cargo-plugin.adoc
OndrejM: While a profile isn't *needed* as @mikecroft  says, it's true that without a profile it's not easy to trigger a specific plugin execution from command line. Before maven v3.3.1, it wasn't possible at all: https://stackoverflow.com/a/28642594/784594

The role of a profile is to wrap a specific execution, e.g. for deploy, redeploy, etc. into a profile. Then a specific action like `deploy` can be triggered just by enabling the profile. On the command line, it would be like: `mvn -Pdeploy package`

So I would rephrase that in order to execute a deploy action from command line (and not only from within a build lifecycle), you really need to create a profile.

OndrejM: I suggest using words like "we will" instead of "you need", which would avoid these issues whether something is required or only a best practice. This is a guide, which means it should show how to do things, not to discuss all possible options. Words "we will" instruct users, but don't claim that the steps are required.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
OndrejM: Another example that using "we will" instead of "you need" is more appropriate and doesn't claim that the steps are necessary.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
OndrejM: @mikecroft is correct. While specifying the cargo property also works, it would bypass the value of `payara.username`. I would encourage using the `payara.*` properties. They're there for consistency because the cargo properties don't have a consistent naming pattern with `cargo.remote.*`, `cargo.glassfish.*`, etc.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
OndrejM: I would use `password` instead of `adminadmin`, because it's more often used for examples. And it should be clarified that this is not a default value but specified by a user as @mikecroft said.

OndrejM: Should be `payara.password` instead.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
```diff
@@ -0,0 +1,161 @@
...
+
+This is a remote profile which relies on using an existing Payara Server
+installation.  You will need to specify the following within the `<properties>`
+tags in your pom:
+
+[source, xml]
+----
+<cargo.remote.username>admin</cargo.remote.username>
+<cargo.remote.password>adminadmin</cargo.remote.password>
+<admin.port>4848</admin.port>
```
OndrejM: Should be `payara.adminPort` instead.

documentation/user-guides/using-the-maven-cargo-plugin.adoc
OndrejM: Should be `payara.hostname` instead. Plain `hostname` wouldn't work, you probably meant `cargo.hostname`, but `payara.hostname` is more appropriate.

mikecroft:
Superceded by #354

MarkWareham:
Looks like this PR should be deleted @OndrejM ?

PR #290, opened 2018-02-01
PAYARA-2203 Document MP Healthcheck

Pandrex247:


documentation/microprofile/healthcheck.adoc
```diff
@@ -0,0 +1,30 @@
+= Eclipse MicroProfile Healthchecks API
```
OndrejM: I'm not sure that Healthchecks is the official spec name. I've raised an issue for MP to clarify the name: https://github.com/eclipse/microprofile-health/issues/118

OndrejM: This is not very important, we can change it later when clarified.

mikecroft: IMO - names of specs are ***very*** sketchy. Healthchecks is the worst, but JWT is bad too - I've seen it as "JWT", "JWT RBAC", "JWT Interop" etc. We need to get names to be clear in MP across the board.

OndrejM:
I think it's clear the this feature is about an application API and REST endpoint, while the Healthcheck service is about self-checks and reporting alerts. Maybe they could even be merged together as a single Healthcheck functionality, with the Healthcheck service being able to send alerts about MP Health status...

PR #293, opened 2018-02-06
PYSVC-639 MicroProfile Metrics

mikecroft:
Will probably need a rebase/conflict fix for the summary due to how close the other MP docs will be

### DIFF
--- a/documentation/microprofile/metrics.adoc
+++ b/documentation/microprofile/metrics.adoc
@@ -2,6 +2,8 @@
 
 _Since *4.1.2.181*; *5.181*&nbsp;_
 
+Provided version of the API: {{ book.mpMetricsSpecUrl }}[MicroProfile Metrics {{ book.mpMetricsVersion }}]
+
 == Background
 The Metrics API was designed to provide monitoring support for applications as well as the underlying runtime. This is considered separate to ***health checks*** which are concerned with a binary status of "healthy" or "unhealthy". In design, the Metrics API has borrowed heavily from DropWizard Metrics, so many of the concepts in the API will be familiar.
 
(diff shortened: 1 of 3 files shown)


documentation/microprofile/metrics.adoc
```diff
@@ -0,0 +1,60 @@
...
++
+These metrics can be accessed under `/metrics/vendor`.
+
+**Application**::
+Application metrics are reserved for those metrics which are to be exposed from the application by using the developer API.
++
+These metrics can be accessed under `/metrics/application`.
+
+== Custom Vendor Metrics
+JMX Mbeans can be exposed as ***Custom Vendor Metrics*** by the user by supplying a custom https://github.com/payara/Payara/blob/f30de88abca76111caadf0d57249d77d53b067e7/appserver/payara-appserver-modules/microprofile/metrics/src/main/resources/metrics.xml[metrics.xml] file in the `$PAYARA_HOME/glassfish/domains/$DOMAIN_NAME/config/` folder with the following format:
```
OndrejM: Do you know if it's possible to add `metrics.xml` also to Payara Micro? This sounds it's nly available in Payara Server, while the NOTE below suggests it might work with Payara Micro too.

mikecroft: Metrics can be added, but not in a nice way, you need to use the `--rootdir` workaround. Also Micro doesn't have any of the same Payara-specific MBeans as Payara Server, so you can only monitor JVM MBeans like `java.lang` or something.

documentation/microprofile/metrics.adoc
```diff
@@ -0,0 +1,60 @@
+= Eclipse MicroProfile Metrics API
```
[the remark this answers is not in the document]
mikecroft: Yes, I thought about that, but it's the same for all specs. The versions are listed in the microprofile parent page, I wasn't sure what the best way to handle that was.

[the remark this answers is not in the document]
OndrejM: I agree that it should be handled in the same way for all specs. I'll try to update all other specs in a similar pattern.

documentation/microprofile/metrics.adoc
OndrejM: That's fair enough.

PR #294, opened 2018-02-09
Page about technical preview

OndrejM:
Adds page general-info/technical-preview.adoc about technical preview.

This page has a Warning dialog that can be cpied to documentation of features in tech preview.

A preview in gitbook: https://docs.payara.fish/v/c3c12277676d1c995b8170e61b64e772ff33683a/general-info/technical-preview.html


SUMMARY.adoc
```diff
@@ -142,7 +143,7 @@
 *** link:documentation/payara-micro/clustering/clustering-with-full-server.adoc[Clustering with Payara Server]
 *** link:documentation/payara-micro/clustering/lite-nodes.adoc[Lite Cluster Members]
 ** link:documentation/payara-micro/maven/maven.adoc[Maven Support]
-** link:documentation/payara-micro/port-autobinding.adoc[HTTP(S) Auto-Binding]
+** link:documentation/payara-micro/port-autobinding.adoc[HTTP\\(S\\) Auto-Binding]
```
Pandrex247: Is that right?

OndrejM: I think this was added by the gitbook editor. I reverted it, thanks.

OndrejM:
@mikecroft  please review.

I found a solution to use both an included fragment and a version condition. The same include statement can be used anywhere as is, just with a different previewVersion.

Unfortunately, it's not possible to nest an include statement in a condition like this:

```
{% if book.currentVersion == "4.1.2.181" %}
    {% include "/fragments/tech-preview.adoc" %}
{% endif %}
```

The include just doesn't work even if the condition is true.

OndrejM:
The current solution is to write the following to include the tech preview fragment:

```
{% set previewVersion = "4.1.2.181" %}
{% include "/fragments/tech-preview.adoc" %}
```

PR #296, opened 2018-02-13
Aligned formatting of code snippet

OndrejM:
As other code snippets


mikecroft:
Most other snippets of this kind are formatted in the other way because they are single line commands, and this style emphasises that. We could change it but, IMO, the current style looks good (particularly for shorter commands).

mikecroft:
Actually, the thing that differs (IIRC) is the example. In other files, we've formatted them as blocks with each option on a new line to make it very clear how many options there are and what the possible values are. That might benefit from a change here.

OndrejM:
You're right, @mikecroft . The example is different but it's different also in other pages so I'll let it be. Closing this PR.

PR #309, opened 2018-02-20
PYSVC-664 document the config secrets dir config source for microprofile

smillidge:


documentation/microprofile/config.adoc
```diff
@@ -27,6 +27,13 @@ supported ConfigSources:
 |Ordinal
 |Description
 
+|`secrets directory`
+|90
+|Reads configuration properties from files in a specified directory. Each file name 
+is the confguration property name and the contents of the file are the config property
+value. This config source can be used to read https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod[Kubernetes secrets files].
```
OndrejM: It can also be used to read Kubernetes Configmap (https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) - they both have the same format.

documentation/microprofile/config.adoc
```diff
@@ -27,6 +27,13 @@ supported ConfigSources:
 |Ordinal
 |Description
 
+|`secrets directory`
```
OndrejM: This feature should be renamed just to "directory" because it's not specific for Kubernetes Secrets.

Moreover, the secrets are base64 encoded but AFAIK this config source doesn't decode them and leaves it for the application. All the asadmin commands and code refers to "secrets" - I'm going to raise a JIRA to propose that those are renamed in the next Payara version.

documentation/microprofile/config.adoc
```diff
@@ -300,3 +309,75 @@ asadmin> get-config-property
     --target=myAppCluster
 ----
 
+==== `set-config-secrets-dir`
+
+*Usage*::
+`asadmin> set-config-secrets-dir --directory=<full.path.to.dir> --target=<target[default:server]>`
+*Aim*::
+Sets the directory to be used for the secrets directory config source.
```
OndrejM: "secrets directory config source" should be renamed to "directory config source" as I explained previously.

documentation/microprofile/config.adoc
OndrejM: Is the value `secrets directory` used in asadmin commands in the same way as `domain`, `jndi`? In that case, how does it work with the space in the value? I though that the first column in this table are arguments that can be used in asadmin commands like set-config-ordinal to refer to the config source...

documentation/microprofile/config.adoc
```diff
@@ -300,3 +309,75 @@ asadmin> get-config-property
...
+
+[cols="1,10,1,1", options="header"]
+|===
+|Option
+|Description
+|Default
+|Mandatory
+
+|`directory`
+|Full path to the directory containing secrets files
```
OndrejM: As described above, it's not only about secrets, so should be "...directory with configuration files"

documentation/microprofile/config.adoc
```diff
@@ -300,3 +309,75 @@ asadmin> get-config-property
...
+
+|===
+
+
+===== Example
+
+[source,Shell]
+----
+asadmin> set-config-secrets-dir
+    --directory=/hoe/payara/.secrets
```
OndrejM: Typo - should be "home" not "hoe"

documentation/microprofile/config.adoc
```diff
@@ -300,3 +309,75 @@ asadmin> get-config-property
...
+    --target=myAppCluster
+----
+
+==== `get-config-secrets-dir`
+
+*Usage*::
+`asadmin> get-config-secrets-dir --target=<target[default:server]>`
+*Aim*::
+Gets the value of the directory to be used for the secrets directory config source.
```
OndrejM: Again since not secrets-specific, I would remove the word "secrets" and just leave "...used for the directory config source"

OndrejM:
The new directory config source should be marked as [technical preview](https://github.com/payara/Payara-Server-Documentation/blob/master/general-info/technical-preview.adoc). If you agree that it should be renamed from secrets directory to just directory, then it's worth to explain this under the tech preview notice because asadmin commands will still have the word "secrets" in them until 182.

smillidge:
Is there a JIRA for a rename?
I don't know how to mark it as tech preview in the docs.

documentation/microprofile/config.adoc
smillidge: This is code that is committed and released therefore that is the command name

smillidge:
The new source should not be mentioned in the other source commands as it doesn't use those commands as it doesn't create, set or delete files.

documentation/microprofile/config.adoc
OndrejM: I didn't mean in `get-config-secrets-dir` but in the sentence below: "...for the **secrets** directory".

OndrejM:
> The new source should not be mentioned in the other source commands as it doesn't use those commands as it doesn't create, set or delete files.
The asadmin command to set ordinal could be used with the new source, couldn't it? If yes, how would the command look like?

OndrejM:
@smillidge 
> Is there a JIRA for a rename?
Not yet. I want to test the usability of the feature and report improvement JIRAs afterwards when I have a bigger picture. For 181 it can remain secrets directory, I just wanted to make it clear in the docs that it can be used also for something else than secrets.

> I don't know how to mark it as tech preview in the docs.
I've created a PR into your branch, you may just merge it: https://github.com/smillidge/Payara-Server-Documentation/pull/1

PR #310, opened 2018-02-20
PYSVC-686 document the delete-config-property asadmin command

smillidge:


documentation/microprofile/config.adoc
```diff
@@ -232,6 +232,76 @@ asadmin> set-config-property
 ----
 
 
+==== `delete-config-property`
+
+*Usage*::
+`asadmin> delete-config-property --propertyName=<property.name> --source=domain|
+config|server|application|module|cluster --sourceName=<source.name>
```
OndrejM: Should also have a value for the new directory config source in the list of values for `--source`

documentation/microprofile/config.adoc
```diff
@@ -232,6 +232,76 @@ asadmin> set-config-property
...
+--moduleName=<module.name> --target=<target[default:server]>`
+*Aim*::
+Deletes the given property name in one of the built-in config sources so that the property no longer exists. The
+source is specified with `--source` and, where there is ambiguity, the `--sourceName`
+and `--moduleName` options can be used. For example, where the source is `server`,
+the `--sourceName` can be used to specify the name of the server where the config
+property is to be stored. moduleName should only be used when the `--source=module`.
```
OndrejM: "moduleName" could `--moduleName` in monospace text

documentation/microprofile/config.adoc
smillidge: no it shouldn't as it doesn't manage the files directly it just reads them and therefore is not used in this command

OndrejM: Good point, I didn't realize that.

PR #311, opened 2018-02-27
PYSVC-664 Note about tech preview

OndrejM:


OndrejM:
@mikecroft Could you please approve?

OndrejM:
This is exactly the goal, @fturizo - all features in tech preview should be reviewed by the next version. If they aren't they will appear is out of tech review.

Obviously, also tech preview for 4.181 should disappear for 5.181, which is undesirable in most cases. Therefore they would have to be updated.

This may be tedious but we'll see how it goes.

PR #318, opened 2018-03-06
PYSVC-681 Document-immediateExit-attribute-of-Payara-Micro-Maven-plugin

mulderbaba:


documentation/ecosystem/maven-plugin.adoc
```diff
@@ -3,8 +3,8 @@
 Payara Micro has a Maven plugin which can start and stop Payara Micro and create
 an Uber JAR bundle of the application and Payara Micro.
 
-The plugin requires JDK 7 or higher and the latest version is currently
-`1.0.0-SNAPSHOT`
+The plugin requires JDK 8 or higher and the latest version is currently
+`1.0.1-SNAPSHOT`
```
OndrejM: It only makes sense to document final versions in maven central. This documentation should be published after v1.0.1 is published.

PR #335, opened 2018-03-13
PYSVC-740

Pandrex247:


general-info/supported-platforms.adoc
```diff
@@ -11,10 +11,9 @@ Similarly, other platforms can always be requested for support.
 
 Payara Server currently supports the following Java Virtual Machines:
 
-* Oracle JDK 7/8
-* Azul Zulu 7/8
-* OpenJDK 7/8
-* IBM J9 7/8
+* Oracle 8u162 and Upwards
```
OndrejM: Should we also mention that we don't support JDK 9 and 10 yet? The text suggests that we support all versions since 8u162 while I think we only support JDK 8 since u162. So better to write "JDK 8 since update 162" rather than "8u162 and Upwards"

PR #338, opened 2018-03-18
PAYARA-2615 documentation for the --sslCert command line switch intro…

smillidge:
…duced in 182 for both 4 and 5.

NOTE this is for 182 4 and 5.


OndrejM:
Accidentally merged into v5.181 but should go to v5.182 docs. I will fix this later

smillidge:
this one is for 182 so should be reverted.

PR #340, opened 2018-03-19
Sync from 4.181

OndrejM:


OndrejM:
This is an invalid PR from 4.181. Master should be synchronized with 5.181: https://github.com/payara/Payara-Server-Documentation/pull/347

PR #351, opened 2018-04-20
Increase MP Metrics version variable

OndrejM:
Improves #350


OndrejM:
This is only an improvement of the current status. If MP Metrics 1.1 won't make it into 182 then we'll have to revert the docs in master.

Cousjava:
Metrics 1.1 have already been merged in payara/payara#2563

PR #359, opened 2018-06-01
PAYARA-2739 Document set-amx-enabled command

MeroRai:


documentation/payara-server/jmx-monitoring-service/amx.adoc
```diff
@@ -0,0 +1,18 @@
+[[boot-amx]]
+= Boot AMX
+
+_Since Payara Server 5.182_
+
+To boot AMX at startup or independently from the JMX Monitoring service, run the 
+command below: 
+[source, shell]
+----
+asadmin>set configs.config.server-config.amx-configuration.enabled=true
```
Cousjava: The command is set-amx-enabled that is being documented

PR #382, opened 2018-06-25
PAYARA-2814 Document contextRoot option for Payara Micro maven plugin

jGauravGupta:


documentation/ecosystem/maven-plugin.adoc
```diff
@@ -61,7 +61,17 @@ Example:
...
+
+|`autoDeployEmptyContextRoot`
+|true
+|Sets context root of the deployed artifact to `/`  if `autoDeployArtifact` is `true`.
+If set to false, context root is derived from `project.build.finalName`.
+'autoDeployContextRoot' option value overrides the `autoDeployEmptyContextRoot`.
+
+|`autoDeployContextRoot`
+|_none_
+|Sets context root of the deployed artifact if `autoDeployArtifact` is `true`.
```
OndrejM: This should also clarify that if specified, it takes precedence over autoDeployEmptyContextRoot, which is ignored in that case - I hope that's true :)

PR #397, opened 2018-07-26
PAYARA-2922 No OpenAPI Target Option

MattGill98:


OndrejM:
Please do not merge if it applies only to 5.183 and not 5.182

MattGill98:
This doesn't correspond to a bug fix, only a documentation mistake. This will apply to 182 and 183.

PR #416, opened 2018-09-06
Prepare release 183

OndrejM:


rdebusscher:
In the release notes
MP 2.0 includes also the addition of JSON-B 1.0 and upgrade of CDI

Important to mention?

OndrejM:
Technically, even REST client 1.1 was already present in 5.182, so I removed it from the 5.183 release notes and added it to 5.182. I also added FT 1.1 to the 5.182 release notes, which was updated in 5.183.

rdebusscher:
Fine for me. 

I just mentioned it because it could be confusing since MP 2.0 has added them, although when you are already Java EE 8 compliant, you are already implementing them.

PR #435, opened 2018-11-27
PAYARA-3279 Document Environment variable replacement in logging.properties

MeroRai:


documentation/payara-server/server-configuration/var-substitution/usage-of-variables.adoc
```diff
@@ -77,3 +77,10 @@ An example of variable references within a pre or post boot command file:
 set configs.config.server-config.http-service.virtual-server.server.default-web-module=${ENV=deployed-app-name}
 set configs.config.server-config.network-config.protocols.protocol.http-listener-1.security-enabled=${ENV=security-enabled}
 ----
+
```
fturizo: Add an admonition to state that logging properties substitution is available from 5.184 onwards, like this:

NOTE: Environment variable substitution in the logging properties file is available since Payara Server 5.184.

PR #461, opened 2018-12-18
Clarify boot-amx usage

rdebusscher:


documentation/user-guides/monitoring/mbeans.adoc
```diff
@@ -161,8 +161,9 @@ for how to view all available MBeans.
 
 IMPORTANT: Payara MBeans are only available when AMX is booted. This is done
 automatically once a JMX client connects. For all other cases, AMX must be either
-configured to boot at server startup or booted via a JMX operation `bootAMX`
-on the `amx-support:type=boot-amx` MBean.
+configured to boot at server startup (through domain config) or booted via a JMX operation `bootAMX`
+on the `amx-support:type=boot-amx` MBean. Programmatic activation during application
```
OndrejM: I would add "on the amx-support... MBean **using a non-JMX client (e.g. over a local socket)** " to stress more that this should be accessed externally via a JMX connection and not from within an application.

rdebusscher: Document updated to include remark.

PR #468, opened 2019-01-31
Document useBundledJsf config

rdebusscher:


documentation/payara-server/classloading.adoc
```diff
@@ -110,6 +110,26 @@ For *EAR* applications, you can include the
 With this, all libraries included on the EAR's `lib/` directory will
 take precedence.
 
+[[use-bundled-jsf]]
+== Use other JSF implementation
+
+In order to make the server to use the bundled JSF implementation within the application, you need to set an additional configuration parameter.
+You need to indicate within the _glassfish-web.xml_ file that the server should use the bundled JSF implementation as follows:
```
MarkWareham: or payara-web.xml

MarkWareham: but that would be different doctype and root element to below

documentation/payara-server/classloading.adoc
```diff
@@ -110,6 +110,26 @@ For *EAR* applications, you can include the
...
+<!DOCTYPE glassfish-application PUBLIC "-//GlassFish.org//DTD GlassFish Application Server 3.1 Java EE Application 6.0//EN" "http://glassfish.org/dtds/glassfish-application_6_0-1.dtd">
+<glassfish-application>
+    ...
+    <classloading-delegate>false</classloading-delegate>
+    <property name="useBundledJsf" value="true" />
+    ...
+</glassfish-application>
+----
+
+By specifying both options, disable class loading delegation and useBundledJsf, the bundled JSF implementation within your deployment (WAR or EAR) will be used correctly.
```
MarkWareham: Just something to make it read a little easier:
change the comma after "options" to a colon or semi colon. It wasn't clear when I read it first time that these were the options you were stating. I thought it was the outcome, so the words seemed a bit muddled.

documentation/payara-server/classloading.adoc
rdebusscher: I was also thinking about it, but the classloading document says 
Since Payara 4.1.1.162

So I find it tricky to put the example than with payara-web.xml. Because then it will not be working in versions other than 5.182+

documentation/payara-server/classloading.adoc
rdebusscher: I will rephrase

documentation/payara-server/classloading.adoc
rdebusscher: I changed now to payara-web.xml in the example (I even used the wrong code in the previous commit due to copy-paste)

documentation/payara-server/classloading.adoc
```diff
@@ -110,6 +110,24 @@ For *EAR* applications, you can include the
 With this, all libraries included on the EAR's `lib/` directory will
 take precedence.
 
+[[use-bundled-jsf]]
+== Use other JSF implementation
+
+In order to make the server to use the bundled JSF implementation within the application, you need to set an additional configuration parameter, class loading delegation alone is not enough.
```
OndroMih: I think grammatically this is better: "In order to make the server use..." (without **to** use)

PR #469, opened 2019-02-05
PAYARA-3462 added section on new health check asadmin commands

jbee:
This adds the documentation changes for https://github.com/payara/Payara/pull/3663 and https://github.com/payara/Payara/pull/3679.

I used AsciidocFX to edit sources and generate the updated HTML.

In general 3 seconds on 3 new commands were added. These are meant to replace several commands in the future that I marked deprecated for now. I used the `INFO` tag for that at the very beginning of the deprecated commands as I found it used elsewhere in the documentation.

The difference to the replaced commands often is marginal. I did this slowly and careful so hopefully this pays off and I got the details correct and did not mix it up. The examples given are based on the examples given for the replaced commands and were adopted to the new commands. For the new commands I also documented some parameters that did exist before but were undocumented.

### DIFF
--- a/documentation/payara-server/health-check-service/asadmin-commands.adoc
+++ b/documentation/payara-server/health-check-service/asadmin-commands.adoc
@@ -9,15 +9,18 @@ to correctly configure the HealthCheck Service.
 
 *Usage*::
 ----
-set-healthcheck-configuration --enabled=true|false --dynamic=true|false  
---trace-enabled=true|false --trace-store-size=20 
---trace-store-timeout=<integer.value>s|m|h|d]
+set-healthcheck-configuration 
+ --enabled=true|false 
+ --dynamic=true|false  
+ --historic-trace-enabled=true|false 
+ --historic-trace-store-size=20 
+ --historic-trace-store-timeout=<integer.value>s|m|h|d
 ----
 
 *Aim*::
 Enables and disables the HealthCheck service. This includes configuration for tracing historic health check events for later inspection. 
 
-[[command-options]]
+[[command-options-8]]
 ==== Command Options
 
 [cols="3,1,5,1,1",options="header"]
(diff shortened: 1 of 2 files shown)


documentation/payara-server/health-check-service/asadmin-commands.adoc
```diff
@@ -4,9 +4,93 @@
...
+
+|`--historic-trace-store-size`
+|Integer
+|Sets the maximum number of health checks to store
+|20
+|no
+
+|`--historic-trace-store-timeout`
+|String
+|Sets the time period after which a historic health check event entry is removed from visable history. The time expression should consist of a number followed by a time unit; `s` for seconds, `m` for minutes, `h` for hours or `d` for days. If no time unit is given the number specifies seconds. If the parameter is zero or unspecified there is no timeout for entries.
```
MarkWareham: I presume it's not possible to combine. Such as "1m10s"

documentation/payara-server/health-check-service/asadmin-commands.adoc
```diff
@@ -4,9 +4,93 @@
...
+|-
+|no
+
+|===
+
+NOTE: Enabling or disabling the health check service in general implicitly also enables or disables the log notifier which is the default notifier. This behaviour is similar to the replaced <<healthcheck-configure>> command.
```
MarkWareham: comma needed after "general" = general, implicitly. Or remove the words "in general"

documentation/payara-server/health-check-service/asadmin-commands.adoc
```diff
@@ -4,9 +4,93 @@
...
+
+[[example-8]]
+==== Example
+
+The following example will enabled the Healthcheck service such that it will
```
MarkWareham: enabled = enable

documentation/payara-server/health-check-service/asadmin-commands.adoc
```diff
@@ -482,9 +813,98 @@ asadmin> healthcheck-stuckthreads-configure
...
+
+|`--enable`
+|Boolean
+|Enables or disables the notifier
+|false
+|Yes
+
+|`--noisy`
+|Boolean
+|Sets the notfier to noisy or not noisy. If not set no change is applied to current setting.
```
MarkWareham: what is noisy?

documentation/payara-server/health-check-service/asadmin-commands.adoc
jbee: Correct. Its `<value><unit>`.

documentation/payara-server/health-check-service/asadmin-commands.adoc
jbee: I was assuming this is similar to what usually is called _verbose_. Had a look and found this:
```java
    public boolean isEnabled(Level level) {
        return Level.FINE.equals(level) ? enabled && noisy : enabled;
    }
```
So effectively its an additional toggle to turn on the `FINE` log level.
I come up with some text to make it clear.

OndroMih: notfier -> notifier

isNoisy isn't a toogle for FINE level. It's used to disable some messages in FINE log level. If FINE logging is enabled, this can be used to ignore some less important log messages to make the logging less "noisy". 

So with noisy = false: and log level FINEST (logs all):

* all messages above FINE level are logged
* some messages with FINE level are logged (those that ignore the noisy configuration)
* other messages with FINE level are not logged (these respect the noisy configuration)
* messages below FINE level are logged

This effectively adds one more logging level between FINE and FINER for notifiers, which can be enabled/disabled by the noisy argument.

jbee: Thanks @OndroMih for clarifying this and the great explanation. I might have expressed myself a bit sloppy but I thought of it as roughly what you said. Let me see if I can improve the text in the docs a bit more.

PR #474, opened 2019-02-28
PAYARA-3534 Document daily rotation

AlanRoth:
Added some basic information about daily log rotation and how to enable it in Payara Server


documentation/payara-server/logging/daily-log-rotation.adoc
```diff
@@ -0,0 +1,25 @@
+[[daily-log-rotation]]
+= Daily Log Rotation
+_Since version 5.184_
```
rdebusscher: This option is already available on older versions. Don't find it immediately which one it was introduced.

documentation/payara-server/logging/daily-log-rotation.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+Payara has a easy to use daily rotation setting in the admin console that will rotate server logs every day at midnight, checking 'Rotation on Date Change' will enable this.
+
+image:/images/logging/daily-log-rotation.png[Daily rotation config]
+
+By default a size rotation of 2mb is used for logs in Payara Server, meaning no log files will be deleted until the size limit is reached and a new log is made at midnight.
+
+Payara Server also has a number of rotation conditions which can be changed.
+
+* Time - Daily, weekly, monthly or even hourly log rotation.
```
rdebusscher: How do you set these?

documentation/payara-server/logging/daily-log-rotation.adoc
```diff
@@ -0,0 +1,25 @@
...
+* Size - Logs are rotated when they exceed a certain limit.
+* Number - Logs are rotated when a maximum number of logs is reached.
```
rdebusscher: `maximum number of logs is reached.` I think this is the number of log entries in the file.

documentation/payara-server/logging/daily-log-rotation.adoc
AlanRoth: Should I include a picture pointing it out or would mentioning that it can be done in the admin console be enough? I'm not sure, but can you use asadmin commands to set these?

documentation/payara-server/logging/daily-log-rotation.adoc
AlanRoth: Ok, I wasn't sure

documentation/payara-server/logging/daily-log-rotation.adoc
AlanRoth: I'll have a look

PR #475, opened 2019-02-28
PAYARA-3531 Updated system-properties.adoc

cubastanley:
Added a new system-properties entry for the JLine logging disable feature when using asadmin multimode

### DIFF
--- a/documentation/payara-server/system-properties.adoc
+++ b/documentation/payara-server/system-properties.adoc
@@ -20,7 +20,7 @@ directory |true, false |true
 
 |fish.payara.permittedFormBasedAuthHttpMethods|String| A comma separated list of HTTP methods that are permitted to use for FORM based authentication (`j_security_check`). If not set all HTTP methods are permitted. | GET,HEAD,POST,PUT,DELETE,CONNECT,OPTIONS,TRACE,PATCH | null (undefined)
 
-|fish.payara.admin.command.jline.log.disable|Boolean| When set to true, JLine logging entries for invalid key bindings and other unwanted errors will be prevented | true, false | false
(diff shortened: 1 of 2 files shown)


documentation/payara-server/system-properties.adoc
```diff
@@ -20,5 +20,7 @@ directory |true, false |true
 
 |fish.payara.permittedFormBasedAuthHttpMethods|String| A comma separated list of HTTP methods that are permitted to use for FORM based authentication (`j_security_check`). If not set all HTTP methods are permitted. | GET,HEAD,POST,PUT,DELETE,CONNECT,OPTIONS,TRACE,PATCH | null (undefined)
 
+|fish.payara.admin.command.jline.log.disable|Boolean| When set to true, JLine logging entries for invalid key bindings and other unwanted errors will be prevented | true, false | false
```
OndroMih: This property isn't for Payara Server but for the asadmin tool. Please make that explicit. To apply the property, the bin/asdamin file (or bin\asadmin.bat on Windows) should be edited to change the last line like this:

```
exec "$JAVA" -Dfish.payara.admin.command.jline.log.disable=true -jar "$AS_INSTALL_LIB/client/appserver-cli.jar" "$@"
```

cubastanley: Will changing that line update the property at runtime or will the server require a restart?

OndroMih: This only affects asadmin, which is a command line tool to send commands to Payara Server. No need to restart Payara Server. This option only disables some irritating logging in the asadmin tool itself.

PR #478, opened 2019-02-28
PAYARA-3533 Document new log property

AlanRoth:


OndroMih:
What about an asadmin command to set this?

documentation/payara-server/logging/date-stamp-log-property.adoc
```diff
@@ -0,0 +1,6 @@
+[[date-stamp-log-property]]
+= Date stamp log property
+
+__Since version 5.184__
```
rdebusscher: since 5.191

documentation/payara-server/logging/date-stamp-log-property.adoc
```diff
@@ -0,0 +1,6 @@
+[[date-stamp-log-property]]
+= Date stamp log property
```
rdebusscher: Make it clear it is for Access Log

documentation/payara-server/logging/date-stamp-log-property.adoc
```diff
@@ -0,0 +1,6 @@
...
+
+__Since version 5.184__
+
+The date stamp for access logs can be configured. Instead of every log file being date stamped, only the first access log file can be date stamped. The date stamp is only added starting with the first rotation.
```
rdebusscher: I find the explanation a bit confusing.

When property is set to true, only the files which are rotated at the end of the day, get the date stamp. When the parameter is false, the file being written is already having the date stamp appended

PR #487, opened 2019-03-06
PAYARA-3525 PAYARA-3526 Document new EJB remote HTTP feature

arjantijms:
PAYARA-3525


OndroMih:
After merging this, we should split the page into multiple pages per section. The page is getting too long.

PR #494, opened 2019-03-07
Fixed release notes version

OndroMih:
to point to 191 instead of 184


OndroMih:
This should be merged to 5.191 too.

PR #501, opened 2019-04-09
NP-42 Update Payara Micro NetBeans Plugin screenshot

jGauravGupta:


MarkWareham:
Please don't forget to include the issue number in title. I see from your branch what it is though

PR #536, opened 2019-06-07
PAYARA-3821 Add asadmin command to reset AMX metrics Docs

AlanRoth:


documentation/payara-server/monitoring-service/asadmin-commands.adoc
```diff
@@ -246,4 +246,80 @@ thread-pool                  OFF
...
+You can use the verbose mode to see which levels were restarted:
+
+[source, shell]
+----
+asadmin > restart-monitoring
+     --target=server
+     --verbose
+----
+
+Verbose mode will show you which modules were affected and what monitoring level they are set to.
```
cubastanley: I feel like this sentence should be merged with the statement before the code block to have something like "You can use the verbose switch to see which modules were affected and what monitoring level they are set to" - would flow better with having the code block followed by its output

AlanRoth: You're right, I was repeating myself

PR #539, opened 2019-06-13
Document removal of System properties

rdebusscher:


OndroMih:
Yesterday we agreed that removal of the ext sys prop is a bug and not intended, so this PR should be reverted

PR #540, opened 2019-06-15
Revert removing ext and endorsed sys props

OndroMih:
We've agreed that removing the sys props for JDK 8 was a mistake and it's a bug that will be fixed. See PAYARA-3931. 

Reverting #539


OndroMih:
Hi @rdebusscher, I'm not able to update the original branch anymore, so I've created a new pull request with updates.

Because JDK 11 support is still only in tech review and JDK 8 is the main supported Java version, I still suggest reverting changes done on the production domain page. This is how it's supposed to be and the behavior in 5.192 is a bug.

But I've updated my changes for the release notes to state that those sys props aren't supported for JDK 11 and that there's also a known issue for JDK 8.

PR #542, opened 2019-06-17
Updated release notes - ext sys prop issue

OndroMih:
We've agreed that removing the sys props for JDK 8 was a mistake and it's a bug that will be fixed. See PAYARA-3931.

Reverting #539.

A follow-up for #540 with updates.


OndroMih:
Closing - wrong branch.

PR #544, opened 2019-06-17
Synchronize from 5.192

OndroMih:


OndroMih:
The conflict is because the same thing was committed into both 5.192 and master and then was changed in 5.192. The version in 5.192 is more up to date and should overwrite the version in master.

Pandrex247:
This has been held open too long, we can't merge from 5.192 into master as both branches have moved on.
Doing a git diff doesn't show anything I can actually see changing that is correct, so closing.
If I'm wrong then please cherry-pick the individual commit and make a new PR.

PR #546, opened 2019-07-10
[PAYARA-3891] - Updated JVM options docs

cubastanley:
Small docs update to show the need for jvm options to be in quotations when using the create-jvm-options asadmin command

### DIFF
--- a/documentation/payara-server/server-configuration/jvm-options.adoc
+++ b/documentation/payara-server/server-configuration/jvm-options.adoc
@@ -34,8 +34,8 @@ You can omit either the minimum or maximum versions by leaving the fields blank.
 [[using-asadmin-commands]]
 === Using Asadmin commands
 
-JVM options can be configured using the 'create-jvm-options' asadmin command
-NOTE: The jvm option key value pair must be enclosed in quotation marks as follows - asadmin create-jvm-options --target=server-config "-Djava.awt.headless\=true"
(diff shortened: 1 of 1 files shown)


documentation/payara-server/server-configuration/jvm-options.adoc
```diff
@@ -51,16 +53,16 @@ NOTE: When using these arguments to set multiple JVM options, the specified vers
...
 ----
 +
 You can specify the range for each JVM option when specifying multiple options too:
 +
 [source, shell]
 ----
-asadmin create-jvm-options --target=server-config -Djava.awt.headless\=true:[|1.8.0u172]-XX\:+UnlockDiagnosticVMOptions:[1.8.0u168|1.8.0u172]-XX\:NewRatio\=2
-----
\ No newline at end of file
+asadmin create-jvm-options --target=server-config "-Djava.awt.headless\=true":"[|1.8.0u172]-XX\:+UnlockDiagnosticVMOptions":"[1.8.0u168|1.8.0u172]-XX\:NewRatio\=2"
```
Pandrex247: Is that correct...? I'll need to check it

cubastanley: Tested it to make sure and that is the format that executes correctly

Pandrex247: Tested and yep this works, but so does not surrounding each option in quotes; you can get away with doing it once:

`asadmin create-jvm-options --target=server-config "-Djava.awt.headless\=true:[|1.8.0u172]-XX\:+UnlockDiagnosticVMOptions:[1.8.0u168|1.8.0u172]-XX\:NewRatio\=2"`

PR #588, opened 2019-08-23
PAYARA-3885 Documentation for Implement Parameterized dynamically resolved role names

jGauravGupta:


documentation/payara-server/public-api/roles-permitted.adoc
```diff
@@ -56,6 +56,50 @@ They are detailed below.
...
+== Dynamic Roles
+
+The `@RolesPermitted` annotation also allows to dynamically specify security roles permitted to access methods in an application
+by using a predefined expression language (EL) as a dynamic role expression that interacts with the business 
+method's parameters annotated with `javax.inject.Named` annotation.
+
+[[usage-example]]
+=== Example
+
+Here�s an example that protects a single method by defining dynamically resolved roles name
```
Pandrex247: This sentence is a little long and hard to follow.
It might be better worded as something like so (bear in mind I'm not 100% familiar with the feature):

`Here's an example that protects a single method by defining a dynamically resolved role name. The `@Named("customer")` annotation is used to define the parameter name and also references the parameter value for dynamic role expression.

documentation/payara-server/public-api/roles-permitted.adoc
```diff
@@ -56,6 +56,50 @@ They are detailed below.
...
+@RolesPermitted("#{self.findRole(customer, product)}")
+public void updateCustomer(@Named("customer") Customer customer, @Named("product") Product product){
+}
+
+public String findRole(Customer customer, Product product) {
+    return customer.type + "_" + product.model
+}
+----
+
+If the parameter name not defined using `Named` annotation and the method have single parameter then `param` is used as default parameter name:
```
Pandrex247: If the parameter name **is** not defined using **the** `**@**Named` annotation and the method **has a** single parameter then `param` is used as **the** default parameter name

PR #606, opened 2019-09-06
PAYARA-4126 Update MP support

Pandrex247:


documentation/microprofile/README.adoc
```diff
@@ -217,6 +217,10 @@ Payara has been part of the MicroProfile project from its inception and, therefo
 
 == Payara Support for MicroProfile
```
OndroMih: Can we also move this whole section "Payara Support for MicroProfile" to the top? It's more important than general information about MP which is at the top currently.

PR #622, opened 2019-09-25
PAYARA-3263 Document how to use an alternative opentracing implementation

Cousjava:


documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,10 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
 === OpenTracing Configuration
 
 You can configure the notifiers and open tracing service by using the link:/documentation/payara-server/request-tracing-service/asadmin-commands.adoc[asadmin commands] and also using the link:/documentation/payara-server/request-tracing-service/configuration.adoc[Administration console].
+
+[[alternative-implementation]]
+== Using Alternate Opentracing Implementations
+
+_Since Payara Server 5.194_
+
+If you wish to use an alternative implementation of OpenTracing, i.e. Jaeger, then the alternative implementation must be proved as a jar in the lib directory of your domain. In it must be a file named io.opentracing.Tracer under META-INF/services, with the name of the class implementing io.opentracing.Tracer. That implmentation must have a no-argument constructor. This will trace all JAX-RS and JAX-WS requests but not servlet or JSF requests. Payara's request tracing service MUST also be enabled. All other configuration within Payara's request tracing service will have no effect on the supplied tracer, which must be self-configured.
```
OndroMih: The statement `This will trace all JAX-RS and JAX-WS requests but not servlet or JSF requests` is suspicious and may raise concerns. Does that mean that the native Payara tracer supports also servlet and JSF requests or none of them are supported in any way? Why don't we support them? Is this just because it's not implemented yet or because there are some technical reasons not to implement them?

A little clarification would help. E.g. use the word "yet" like "but not yet servlet or JSF requests". Or that they are not meant to be supported and are out of scope for now. Something that gives a clue to those who ask "why".

fturizo: I also agree with @OndroMih's comment, as it is not clear entirely on to why both Servlet and JSF requests are not traced.

rdebusscher: Since it just loads a Tracer through ServiceLoader mechanism, it doesn't make any difference which one is used regarding the functionality. So I guess, Servlets and JSF requests are never traced (although I seem to remember an example we have that does trace Servlets and CDI calls - if correctly annotated)
So indeed, more explanation required.

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,12 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+[[alternative-implementation]]
+== Using Alternate Opentracing Implementations
+
+_Since Payara Server 5.194_
+
+If you wish to use an alternative implementation of OpenTracing, i.e. Jaeger, then the alternative implementation must be proved as a jar in the lib directory of your domain. In it must be a file named io.opentracing.Tracer under META-INF/services, with the name of the class implementing io.opentracing.Tracer. That implementation must have a no-argument constructor. Alternate Opentracing implementations will only trace JAX-RS, JAX-WS and classes or methods annotated with `@Traced; Payara's own request tracing service can trace link:/documentation/payara-server/request-tracing-service/request-tracing-service.adoc[more types of request]. Payara's request tracing service MUST also be enabled, whichever implementation is being used in order for tracing to occur. All other configuration within Payara's request tracing service will have no effect on the supplied tracer, which must be self-configured.
```
Pandrex247: I think you mean _provided as a jar_, not _proved as a jar_ :P

Pandrex247: `All other configuration within Payara's request tracing service will have no effect on the supplied tracer, which must be self-configured.`

Does this mean that the notification system also stops working? Or does this only apply to the sampling options?

fturizo: This statement needs further clarification. What does mean to be "self-configured"?

Cousjava: "self-configured" means the library has to set all of its configuration inside itself, Payara can't do it.

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
```
rdebusscher: I would put class and directory names in italic.

rdebusscher: missing ` after the @Traced.

documentation/microprofile/opentracing.adoc
rdebusscher: >configuration inside itself

You can configure most of the tracer also through environment or system properties. So doesn't need to be 'inside' the jar.

Cousjava: I still call that self-configured, as it is still done inside the library jar, Payara is not doing any of the configuration itself.

Pandrex247: Personally I'd say it reads better as `which must be configured separately`. 
If you wanted to be really explicit extend it to something along the lines of `which must be configured separately to Payara.`

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+Activating Request Tracing is done either via the admin console or using the commands:
+
+`set-requesttracing-configuration --enabled=true --dynamic=true`
+`bootstrap-requesttracing`
+
+However all other options for request tracing are ignores as the created Spans are not processed within Payara.
+
+As an example, see https://github.com/payara/ecosystem-jaeger-tracing, which shows how to integrate a Jaeger client tracer. This can be added as a library to Payara with the command `add-library jaeger-tracer-lib`. Enabling request tracing in Payara Server is then required, which will then send traces from OpenTracing to a Jaeger collector on the same host.
```
Pandrex247: Wasn't it `jaeger-tracer-lib-with-dependencies`?

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+[[alternative-implementation]]
+== Using Alternate Opentracing Implementations
+
+_Since Payara Server 5.194_
+
+If you wish to use an alternative implementation of OpenTracing, i.e. Jaeger, then the alternative implementation must be proved as a jar in the lib directory of your domain. In it must be a file named _io.opentracing.Tracer_ under _META-INF/services_, with the name of the class implementing _io.opentracing.Tracer_. That implementation must have a no-argument constructor. Alternate Opentracing implementations will only trace JAX-RS, JAX-WS and classes or methods annotated with `@Traced`; Payara's own request tracing service can trace link:/documentation/payara-server/request-tracing-service/request-tracing-service.adoc[more types of request]. Payara's request tracing service MUST also be enabled, whichever implementation is being used in order for tracing to occur. All other configuration within Payara's request tracing service will have no effect on the supplied tracer, which must be configured separately to Payara.
```
rdebusscher: proved -> provided

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+Activating Request Tracing is done either via the admin console or using the commands:
+
+`set-requesttracing-configuration --enabled=true --dynamic=true`
+`bootstrap-requesttracing`
+
+However all other options for request tracing are ignores as the created Spans are not processed within Payara.
```
rdebusscher: ignores -> ignored

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+Activating Request Tracing is done either via the admin console or using the commands:
+
+`set-requesttracing-configuration --enabled=true --dynamic=true`
+`bootstrap-requesttracing`
+
+However all other options for request tracing are ignored as the created Spans are not processed within Payara.
```
fturizo: This sentence would work better as an INFO or WARNING admonition

documentation/microprofile/opentracing.adoc
```diff
@@ -35,3 +35,19 @@ The Payara Platform's MicroProfile implementation supports tracing of requests t
...
+
+[[alternative-implementation]]
+== Using Alternate Opentracing Implementations
+
+_Since Payara Server 5.194_
+
+If you wish to use an alternative implementation of OpenTracing, i.e. Jaeger, then the alternative implementation must be provided as a jar in the lib directory of your domain. In it must be a file named _io.opentracing.Tracer_ under _META-INF/services_, with the name of the class implementing _io.opentracing.Tracer_. That implementation must have a no-argument constructor. Alternate Opentracing implementations will only trace JAX-RS, JAX-WS and classes or methods annotated with `@Traced`; Payara's own request tracing service can trace link:/documentation/payara-server/request-tracing-service/request-tracing-service.adoc[more types of request]. Payara's request tracing service MUST also be enabled, whichever implementation is being used in order for tracing to occur. All other configuration within Payara's request tracing service will have no effect on the supplied tracer, which must be configured separately to Payara.
```
fturizo: Just a small minor format change needed: Class names should be formatted with monospace like this: `io.opentracing.Tracer`

PR #641, opened 2019-10-22
PAYARA-4233 Document CORS headers 5.192

MattGill98:
Maintenance PR of https://github.com/payara/Payara-Server-Documentation/pull/639


documentation/microprofile/openapi.adoc
```diff
@@ -171,6 +173,14 @@ NOTE: When the OpenAPI service is disabled, the `/openapi` endpoint will always
 |true
 |No
 
+|`corsheaders`
+|Boolean
```
OndroMih: This line breaks table formatting. I guess the column for Type was added in a later version.

MattGill98: Good catch!

PR #645, opened 2019-10-23
Removed multiblock comments

OndroMih:
There should be empty values instead, as they were in the original markdown document: https://github.com/payara/Payara-Server-Documentation/blob/ccef1099308ad17789641b0293ca84596e187d73/documentation/extended-documentation/advanced-jdbc/advanced-connection-pool-properties.md. This probably happened during automatic transformation to Asciidoc.


OndroMih:
I'm just fixing the problem. We should fill something in with a separate PR.

PR #651, opened 2019-11-04
PAYARA-4099 Update of docker-instances after testing

dmatej:


documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+
+- payaraDas and dockerHost1 are hostnames known to local naming service (or you can use IP addresses)
+- DAS is listening on port 4848 on payaraDas
+- docker is listening on port 2376 on dockerHost1
+- DAS can access http://dockerHost1:2376
+- docker containers can access https://payaraDas:4848
+- there are no docker containers yet, no Docker nodes and no instances registered in DAS
+- in /app/passwordfile.txt is at least one line with a correct password: AS_ADMIN_PASSWORD=admin123
+
+You should also use some final docker image instead of payara/server-node:latest, because this image changes
```
Pandrex247: This probably isn't necessary.
You'd only be getting `server-node:latest` if you've built the image yourself - the default image is the same as the payara version (e.g. 5.193, or 5.194-SNAPSHOT)

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+IMPORTANT: Please note that it is required that secure-admin be enabled for Docker instances to start
+(which is why a passwordfile is mandatory when creating a Docker node).
 
 [[creating-an-instance]]
 === Creating an Instance
 
-Creating a Docker instance is done in exactly the same way as when creating a normal one: with the create-instance
-command.
+There are two ways how to create a Payara Docker instance belonging to a Domain Administration Server (DAS).
+You can create a Docker container node and instances on it using DAS, but you can also create
```
Pandrex247: `using DAS` should probably have a "the" in there: `using the DAS`

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+
+==== From Domain Administration Server
+
+Creating a Docker instance is done in exactly the same way as when creating a normal one.
+You will create a Docker node and then you can create one or more Payara instances on it.
 
 [source, shell]
 ----
-asadmin create-instance --node DockerNode1 dockerInstance1
+>> asadmin --passwordfile /app/passwordfile.txt create-node-docker --nodehost dockerHost1 --dockerpasswordfile /app/passwordfile.txt --dockerimage payara/server-node:mytag --dockerport 2376 DockerNode1
```
Pandrex247: The example isn't wrong, but it's potentially confusing that the `passwordfile` and `dockerpasswordfile` options are the same.

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+instance names - it will not attempt to resolve conflicts in container name.
+
+WARNING: This type of instance is completely managed by the DAS - when you delete it, it will
+be deleted also the docker container including it's logs.
+
+==== From Docker
+
+Payara Server supports creating instances from Docker itself, making use of the
+link:documentation/payara-server/asadmin-commands/auto-naming.adoc[auto-naming] feature to resolve any conflicts.
+The docker command creates a Payara Docker container without any contact with the DAS.
```
Pandrex247: This is a little misleading.
Yes it creates the Docker container without contact with the DAS, but it still tries to contact the DAS as the container starts when creating the isntance etc.

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+When the Docker container is started for the first time, a temporary Docker node and instance
+will be created on the DAS unless you specify name of existing Docker node in PAYARA_NODE_NAME property.
+The temporary Docker node is bound to the instance, so when you delete the instance,
+the temporary Docker node will be deleted too.
+This type of node is also not listed in the DAS or offered in inputs and cannot be edited.
 
-NOTE: When autonaming is enabled, Payara Server will only attempt to resolve conflicts with instance names - it will not
-attempt to resolve conflicts in container name.
+If the instance is stopped from the DAS, it is immediately unregistered from the DAS.
+Next start of the same Docker container creates new a Payara Instance and a new temporary
```
Pandrex247: Also probably needs a "the": `The next start..."

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+Docker node again despite the fact that it uses the same Docker container.
+
+If the Docker containers are stopped externally to Payara (ie. by the docker command), the node
+and instance will be marked for deletion and cleaned up on shutdown.
```
Pandrex247: `shutdown of the DAS`

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+
+When you set the PAYARA_NODE_NAME and ideally also DOCKER_CONTAINER_IP to an existing Docker node,
+the new container will be managed as if it would be created from DAS.
```
Pandrex247: `from **the** DAS`

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+Command stop-instance executed successfully.
 ----
 
 [[deleting-an-instance]]
 === Deleting an Instance
 
-Much as with when creating a Docker instance, deleting a Docker instance is done in the same way as other instances:
-with the delete-instance command. This will unregister the instance from the DAS, and delete the Docker container.
+Much as with when creating a standard Docker instance, deleting a Docker instance is done
+in the same way as other instances: with the asadmin delete-instance command.
```
Pandrex247: Code block ticks around the commands please

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
 
-A Docker instance is the term used to refer to an instance created on a Docker node. These instances exist within their
-own Docker containers, with the lifecycle of these containers being tied to the instance it was created for.
+A Docker instance is the term used to refer to an instance created on a Docker node.
+These instances exist within their own Docker containers, with the lifecycle of Docker containers
+being tied to the instance it was created for.
+
+Note about prerequisities of examples used in this part:
+
+- payaraDas and dockerHost1 are hostnames known to local naming service (or you can use IP addresses)
```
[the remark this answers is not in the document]
dmatej: I humbly disagree - the reason is that then you have to define those variables and we are moving from readable examples to "something like scripting". Now it provides something user must understand and ensure whenever he would want to use this feature.

Users can create own scripts, but I believe here are better only simple descriptive names ... but italics are good idea, thanks!

documentation/payara-server/docker/docker-instances.adoc
dmatej: Latest is not a snapshot equivalent, I had problems with latest - I don't know what was causing that but latest was overwritten by old version "randomly" from the internet again and again. Can you check if it is not caused by some DAS action?

https://hub.docker.com/r/payara/server-node/tags - seems latest is simply latest deployed ...

See link, why latest is not good idea especially on production ... https://vsupalov.com/docker-latest-tag/

By default it uses for 5.194-SNAPSHOT, which does not exist. You're right that on released version it should not be a problem. I will update those examples to use defaults by default and retagging may stay only as some less important warning :-)

documentation/payara-server/docker/docker-instances.adoc
dmatej: True is that I separated payaraDas and dockerNodeHost in examples, so I should separate also those files, despite I use docker and das on localhost :)
OK, I will fix it.

documentation/payara-server/docker/docker-instances.adoc
dmatej: Yes, but that is next step. It surprised me when I was starting to test, so I would use the same tempo for first-time users.

documentation/payara-server/docker/docker-instances.adoc
```diff
@@ -1,60 +1,184 @@
...
+- DAS is listening on port 4848 on payaraDas
+- docker is listening on port 2376 on dockerHost1
```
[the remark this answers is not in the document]
dmatej: Docker and `docker` command, you're right!

documentation/payara-server/docker/docker-instances.adoc
dmatej: I changed the text perhaps to be more comprehensible, but I would need more distance ... is it better?

PR #657, opened 2019-11-14
PAYARA-4165 Documentation should indicate TX and Security Propagation is Not Supported

AlanRoth:


AlanRoth:
> 
> 
> Does the job but could use an 'Important' warning message format from adoc instead? https://asciidoctor.org/docs/asciidoc-writers-guide/ - scroll to admonitions section

You've confused me, I'm already using the important admonition

cubastanley:
> You've confused me, I'm already using the important admonition

Oh ye, my bad - blame github's lack of rendering, shall approve

PR #650, opened 2019-10-25
PAYARA-3658 Add support for easily securing the ejb-invoker endpoint

jGauravGupta:

PR #659, opened 2019-11-19
Improvements for Lite Remote EJB Client Code

fturizo:
Clarifications were done on how to correctly call secured remote EJBs when using the `ejb-invoker` exposed by the Lite EJB client dependency.

### DIFF
--- a/documentation/payara-server/ejb/lite-remote-ejb.adoc
+++ b/documentation/payara-server/ejb/lite-remote-ejb.adoc
@@ -108,7 +108,7 @@ public interface BeanRemote {
 }
 ----
 
-. Second, consider a (secured) EJB that implements that interface and resides in a EJB application called "test" deployed on a Payara server instance that is listening in `https://localhost:8080`:
+. Second, consider a (secured) EJB that implements that interface and resides in a EJB application called "test" deployed on a Payara server instance that is listening in `http://localhost:8080`:
 +
 [source, java]
 ----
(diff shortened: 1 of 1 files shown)


jGauravGupta:
LGTM. 
Just for clarification with recent improvements in `ejb-invoker` service:

To access the secured `EJB` following properties need to be set from the client:
- `javax.naming.Context.SECURITY_PRINCIPAL`
- `javax.naming.Context.SECURITY_CREDENTIALS`

And to access the secured `ejb-invoker` endpoint following properties need to be set:
- `fish.payara.provider.principal`
- `fish.payara.provider.credentials`

fturizo:
@jGauravGupta, are the properties that you referenced mandatory as well? When I tested the `ejb-invoker` I didn't need them for authentication at all, and the default JNDI properties were enough. Are they replacements?

jGauravGupta:
Hi @fturizo,

Neither these (`fish.payara.provider.principal` & `fish.payara.provider.credentials`) properties are mandatory nor replacement. These properties are required only after enabling the security of `ejb-invoker` endpoint using `asadmin` command: 
`set-ejb-invoker-configuration --enabled=true --securityenabled=true`

See the documentation PR for more detail about the security of `ejb-invoker` endpoint: https://github.com/payara/Payara-Server-Documentation/pull/650

fturizo:
I see, @jGauravGupta. It seems that the new feature that allows the `/ejb-invoker` endpoint to be secured needs these properties. However, I have multiple questions:

* If you configure the ejb-invoker to be secured with the `--securityenabled=true`, are the configurations for transport confidentiality and integrity that are specified in the `glassfish-ejb-jar.xml` deployment descriptor ignored? Will these force all remote EJB calls to be made through HTTPS instead? If this is the case, this needs to be properly documented. 
* I do not see any way to configure the network listener (or virtual servers) to be used by the invoker in the new arguments that were added. Will enabling the invoker to be secure set it under the default `http-listener-2` default secure listener?
* Again, I see that the security configuration for the ejb-invoker allows to set up the list of valid roles, realm, authorization module, etc. Will these settings override EJB specific settings defined in application-level deployment descriptors?
* Why were these properties implemented if there are already standard JNDI properties to send the user credentials in the calls as well? Are these properties exclusive to some conditions? 

It seems that the documentation needs to be expanded to clarify these details.

jGauravGupta:
Hi @fturizo,

- As all lite remote ejb operations performed via `ejb-invoker` endpoint proxy. So enabling security for `ejb-invoker` endpoint does not have any impact on any EJBs or `glassfish-ejb-jar.xml` deployment descriptor but following security configuration [dynamically added](https://github.com/payara/Payara/blob/1845f18cca5f6fb4bf1b7e39f530514d6e692632/appserver/ejb/ejb-http-remoting/endpoint/src/main/java/fish/payara/ejb/invoke/EjbInvokerContainerInitializer.java#L84) to `web.xml` of ejb-invoker client application which enforce the call via `HTTPS`.
````
<security-constraint>
    <web-resource-collection>
        <web-resource-name>EjbOverHttpApplication</web-resource-name>
        <url-pattern>/ejb-invoker/*</url-pattern>
    </web-resource-collection>
    <user-data-constraint>
        <transport-guarantee>CONFIDENTIAL</transport-guarantee>
    </user-data-constraint>
</security-constraint>
````

- Virtual Server can be configured using `--virtualservers` command option. Thanks for the catch, I will update the doc.

- [Ejb Invoker](https://github.com/payara/Payara/tree/9abf9ce49a713d948f004637ea76a3bbbb62b2db/appserver/ejb/ejb-http-remoting/client) is a separate client application. Any security settings does not override EJB specific settings defined in application-level deployment descriptors.

- The `ejb-invoker` endpoint security configuration is very similar to MicroProfile endpoint `/metrics`, `/health`, and `/openapi` security configuration with additionally implemented properties to access the secured `ejb-invoker` endpoint from the client.
The `ejb-invoker` endpoint and EJBs can have different auth configuration (realm, auth module etc) hence standard JNDI properties are not used in this case. 

Let me know If more clarification needed.

PR #686, opened 2019-12-18
Document 'deployment.resource.validation' jvm option.

rdebusscher:


documentation/payara-server/system-properties.adoc
```diff
@@ -25,4 +25,6 @@ directory | true, false | true
...
 
 | org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER | String | If the JVM option `org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER` is not present in the server/instance config, we will override the Grizzly default and use the _HeapMemoryManager_ class. To override this behaviour (so to use a different memory manager), you must specify this JVM option in the server/instance config with your desired setting. | `org.glassfish.grizzly.memory.HeapMemoryManager`, `org.glassfish.grizzly.memory.PooledMemoryManager`, `org.glassfish.grizzly.memory.ByteBufferManager` | `org.glassfish.grizzly.memory.HeapMemoryManager`
+
+| deployment.resource.validation | boolean | This JVM option defines if the resources are validated during deployment of the application. When active, the default value, all JNDI references are verified but those references can be provided by the application itself and thus not yet available at deployment time (for example the `java:module/env/` based ones). By setting this option to false, validation is not performed and deployment will succeed. However, issues will occur at runtime of the application in case of errors in the references.
```
OndroMih: This is not a JVM option but a system property. It doesn't have to be specified as a JVM option. It would work if specified either with the -D JVM option or as a system property on the server instance.

documentation/payara-server/system-properties.adoc
```diff
@@ -25,4 +25,6 @@ directory | true, false | true
 | fish.payara.SO_KEEPALIVE | String | Enables keep alive (`SO_KEEPALIVE`) on the sockets created by an IIOP listener. Can be set as a global system property, or as a property on a specific listener (with the latter taking precedence over the global setting if both are set). Replaces deprecated property `fish.payara.SOKeepAlive`. | true, false | false (undefined)
 
 | org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER | String | If the JVM option `org.glassfish.grizzly.DEFAULT_MEMORY_MANAGER` is not present in the server/instance config, we will override the Grizzly default and use the _HeapMemoryManager_ class. To override this behaviour (so to use a different memory manager), you must specify this JVM option in the server/instance config with your desired setting. | `org.glassfish.grizzly.memory.HeapMemoryManager`, `org.glassfish.grizzly.memory.PooledMemoryManager`, `org.glassfish.grizzly.memory.ByteBufferManager` | `org.glassfish.grizzly.memory.HeapMemoryManager`
```
OndroMih: Please also fix here, this also isn't a JVM option but a system property. However, it's possible that this has to be specified as a JVM option with -D to be applied and wouldn't work if set as server system property as Grizzly is possibly loaded before the server system properties are applied. At least Payara Server sets this property in the launcher with the -D system property, while Payara Micro sets this during runtime from payara-boot.properties before Grizzly is loaded.

documentation/payara-server/system-properties.adoc
rdebusscher: updated, please see updated file with indication of JVM Option with -D or System Property feature.

PR #703, opened 2020-01-27
PAYARA-3793 Documentation for certificate CN and group to role mapping

dmatej:
- how to use CN of DN for principal name
- how to use parts of DN for group names
- how to use standard principal/group names to roles mapping


OndroMih:
Can please clarify what happens if `dn-parts-used-for-groups` contains 2 or more identifiers? Will than a certificate match the mapping if any of those 2 or more parts match the value in any of the `group` element?

E.g., will `C=UK, S=lak, L=zak, OU=unitA, CN=foo1` with ``dn-parts-used-for-groups=OU,CN` match the following rules and assign it `role1` and `foo` roles?

```
<security-role-mapping>
    <role-name>role1</role-name>
    <group-name>unitA</group-name>
    <group-name>unitC</group-name>
    <group-name>unitD</group-name>
  </security-role-mapping>
```

```
<security-role-mapping>
    <role-name>foo</role-name>
    <group-name>foo1</group-name>
  </security-role-mapping>
```

Or all of them have to match? So it would match only the following mapping to assign `role1` role?

```
<security-role-mapping>
    <role-name>role1</role-name>
    <group-name>unitA</group-name>
    <group-name>foo1</group-name>
  </security-role-mapping>
```

dmatej:
It behaves in standard way, if any principal's group matches any group-name OR principal's name matches any principal-name, the principal has the role.

Same with OIDs - all ids from dn-parts-used-for-groups are used to get groups. Before this, groups could be assigned only constantly in realm properties via `assign-groups` property. This property is still respected.

Ths behavior is not explicitly mentioned here, but it is compared to owning keys, so whoever has the key, has the role, regardless it is a uniquie person or a person from some group.
https://javaee.github.io/tutorial/security-intro005.html

I will mention this in documentation now ...

documentation/payara-server/server-configuration/security/security-realms.adoc
```diff
@@ -193,3 +193,15 @@ Then the role mapping can be based on organizational unit in this case. But you
   </security-role-mapping>
 </payara-web-app>
 ----
+
+You have to use the `dn-parts-used-for-groups` property wisely - the set of values must be predictable to be useful.
```
OndroMih: This doesn't clarify very well how the matching works, only gives an example of a conflict. Can you also explicitly describe that the groups in the XML will be matched with all the DN parts specified by the `ds-parts-used-for-groups`? This isn't clearly stated anywhere. Then it makes sense to explain that this can sometimes lead to conflicts.

documentation/payara-server/server-configuration/security/security-realms.adoc
```diff
@@ -74,8 +77,131 @@ Then the `principal-name` can be simplified by using only the CN part instead:
...
+
+[[configure-dn-parts-used-for-groups]]
+== Configure Using Distinguished Name Parts for Groups
+
+_Since Payara Server 5.194_
+
+You can also use parts of client certificate's distinguished name for mapping into groups.
+
+A property named `dn-parts-used-for-groups` can be used with any security realm of type `com.sun.enterprise.security.auth.realm.certificate.CertificateRealm` in order to control this feature.
+This property can contain a comma separated list of identifiers of DN parts, ie. `EMAILADDRESS,DC,OU`
```
OndroMih: Please mention here how the matching works if multiple identifiers are used. This is the place where it should be explained that a rule matches when any value from the group elements is equal to the value of any DN part listed in the property.

dmatej: I expect this should be documented somewhere else, because this is not a feature of realms (see RolesAllowedRequestFilter, ie). Realm work is finished with preconfigured principal name and list of his groups.
But yes, good examples do not raise questions, so I will mention it :-)

PR #724, opened 2020-03-24
APPSERV-20 Document new alias config source

Pandrex247:
Original PR: https://github.com/payara/Payara/pull/4584


documentation/microprofile/config.adoc
```diff
@@ -81,6 +81,14 @@ These are modifiable by current admin console functionality.
 |Retrieves and stores config values in the password alias store. These will be 
 available to all server instances in a domain.
 
+|`alias`
```
OndroMih: Why is this called "alias" config source? It doesn't support only Payara aliases, but all other variable replacements like env variables or sys props. It should rather be called "variables" config source.

documentation/microprofile/config.adoc
```diff
@@ -81,6 +81,14 @@ These are modifiable by current admin console functionality.
...
+|170
+|Reads configuration properties from a microprofile-config-alias.properties file. This
+should be stored in the same location as your microprofile-config.properties file.
+This file can be used to reference aliased properties which will be translated to their
```
OndroMih: To synchronize the terminology, I suggest that we replace "aliased properties" with "variables". In Payara terminology, aliases are only specific variables and it's a subset of what's supported by this config source.

documentation/microprofile/config.adoc
Pandrex247: There wasn't a full discussion had on the naming of the config source and/or file name. I'll throw it out to the team.

Personally I'm not a fan of _variables_.

PR #725, opened 2020-03-26
APPSERV-58 MP Config 1.4

Pandrex247:
Also adds missing table rows for MP 3.1, 3.2, and 3.3


Pandrex247:
There are no functional changes from the user side in Config 1.4.
The only things of note was the addition of Short as a built-in converter.

documentation/microprofile/README.adoc
```diff
@@ -11,6 +11,18 @@ Payara has been part of the MicroProfile project from its inception and, therefo
...
 
+Version 3.3::
+Version 3.3 of MicroProfile updates Config, Health Check, Fault Tolerance, Metrics, and Rest Client.
+This is supported by both Payara Server and Payara Micro from release *5.202*.
+
+Version 3.2::
+Version 3.2 of MicroProfile is a minor release that updates Metrics to maintain backwards compatibility.
+This is supported by both Payara Server and Payara Micro from release *5.194*.
+
+Version 3.1::
```
rdebusscher: MP 3.1 should not be used as it contains breaking changes (and the semantic versioning is thus incorrect). 

The breaking change is reverted in MP 3.2

PR #730, opened 2020-04-07
APPSERV-141 Adds section on added caching and asadmin commands to configure it

jbee:
In connection with https://github.com/payara/Payara/pull/4614

### DIFF
--- a/documentation/microprofile/config.adoc
+++ b/documentation/microprofile/config.adoc
@@ -32,7 +32,7 @@ supported ConfigSources:
 
 |`directory`
 |90
-|Reads configuration properties from files in a specified directory. Each file name 
+|Reads configuration properties from files in a specified directory. Each file name
 is the confguration property name and the contents of the file are the config property
 value. This config source can be used to read https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod[Kubernetes secrets files].
 See below on how to specify the directory to be used for secrets files using asadmin.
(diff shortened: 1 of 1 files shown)


documentation/microprofile/config.adoc
```diff
@@ -578,3 +578,63 @@ asadmin> get-config-secrets-dir
...
 
+
+[[config-cache-duration-configuration]]
+== Config Cache Duration
+
+Config cache duration can be configured by using Asadmin commands.
+By default properties have a TTL (time to live) of 60 seconds.
+That means each individual property does not change for 60 seconds since it has been resolved before.
+Therefore it can take up to 60 seconds for changes made visiable by a config source to become effective.
+In many instances, in particular for properties not resolved regularly change become visible immediate or faster than the cache duration as they have not been resolved for some seconds already.
```
OndroMih, line 589: I can guess what this sentence means but it's not easy to understand.

jbee: What about (bold to highlight change):
_In many instances, in particular for properties not resolved regularly change become visible immediate or faster than the cache duration as they have not been resolved **recently**._

OndroMih: I mean the whole sentence doesn't make much sense, at least grammatically. Maybe something like this would make more sense:

>  If properties were not resolved recently, the change can become visible faster than the cache duration or even immediately because some time already passed since they had been last resolved and cached.

jbee, line 589:
```suggestion
If properties were not resolved recently, the change can become visible faster than the cache duration or even immediately because some time already passed since they had been last resolved and cached.
```

jbee: Thanks, changed it to your suggestion. BTW this would be case where using suggestions comes in handy. Please re-approve ;)

